PR #1, opened 2026-06-28
Add Samsung ARTIK051_REF_17K fridge-freezer

aminorjourney:
This adds local CoAP-DTLS support for Samsung fridge-freezers running the ARTIK051 firmware (`DA-REF-ART-COMMON-1_20201124`), the first known public documentation of this firmware's local resource layout.

**Tested against:** ARTIK051_REF_17K, serial `0B2J43AMXXXXXXH`, firmware `DA-REF-ART-COMMON-1_20201124`, port 49155.

## Key findings vs newer Tizen RT firmware

- `/oic/res` only advertises 15 paths — full resource tree is at `/device/0` (32 links)
- Port is **49155**, not 49154
- `/hass/state/vs/0` and `/hass/command/vs/0` are vestigial (404)
- Energy monitoring present: `/energy/consumption/vs/0` (instantaneous W + cumulative Wh)
- Three doors via `/doors/vs/0`: fridge, freezer, convertible zone

## Entities

| Type | Entities |
|---|---|
| Sensor | Fridge temp, freezer temp, fridge setpoint, freezer setpoint, power (W), energy (kWh), water filter usage %, filter status, ice maker state, ice making status, active modes, serial |
| Binary sensor | Fridge door, freezer door, convertible zone door, any door open, filter OK, ice maker on, power cool, power freeze, sabbath mode |
| Switch | Power cool, power freeze, sabbath mode, ice maker |
| Number | Fridge setpoint (1–7°C), freezer setpoint (-23 to -15°C) |

### DIFF
--- a/bootstrap.py
+++ b/bootstrap.py
@@ -34,7 +34,7 @@ try:
 except ImportError:
     sys.exit("cbor2 not installed — pip install -r requirements-bootstrap.txt")
 
-from samsung_dryer.coap import (
+from samsung_appliance.coap_dtls import (
     URI_PATH, CSM, enc_opts, enc_tcp, read_tcp, fmt_code,
 )
 
--- a/samsung_appliance/appliances/__init__.py
+++ b/samsung_appliance/appliances/__init__.py
@@ -15,6 +15,7 @@ from .oven import OVEN
 from .fridge import FRIDGE
 
 
+
 DESCRIPTORS: dict[str, ApplianceDescriptor] = {
     DRYER.name: DRYER,
     OVEN.name:  OVEN,
--- a/samsung_appliance/appliances/__init__.py
+++ b/samsung_appliance/appliances/__init__.py
@@ -12,11 +12,13 @@ startup; the bridge itself stays class-agnostic.
 from .base import ApplianceDescriptor
 from .dryer import DRYER
 from .oven import OVEN
+from .fridge import FRIDGE
 
 
 DESCRIPTORS: dict[str, ApplianceDescriptor] = {
     DRYER.name: DRYER,
     OVEN.name:  OVEN,
+    FRIDGE.name: FRIDGE,
 }
 
 
--- a/samsung_appliance/appliances/fridge.py
+++ b/samsung_appliance/appliances/fridge.py
@@ -1,0 +1,493 @@
+"""Fridge descriptor — Samsung ARTIK051_REF_17K (and compatible).
+
+Firmware: DA-REF-ART-COMMON-1_20201124
+Protocol: CoAP-DTLS UDP port 49155
+Device ID example: 5ce4dafb-f37e-69f9-8d80-b79def93e140
+
+Resource map discovered 2026-06-27 by Nikki Gordon-Bloomfield
+(github.com/aminorjourney/SmartThings-Local) against a live
+ARTIK051_REF_17K unit — the first known public documentation of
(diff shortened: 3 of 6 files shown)


aminorjourney:
Pushed additional commits beyond the original fridge descriptor — while testing against a real ARTIK051_REF_17K unit over several hours, I found the DTLS session has a variable TTL (observed 8–50 min) and the bridge wasn't recovering from it: polls would silently fail forever, and the keepalive's `on_unreachable` callback could fire *after* a reconnect and falsely flip HA availability offline.

**Changes beyond the fridge descriptor:**
- `publish_health` now detects an all-polls-failing window and triggers a clean session close/reconnect via a per-session stop event
- `KeepaliveTask`'s `on_unreachable` callback is nulled during reconnect so the dying session's keepalive thread can't flip HA availability offline after the new session is already healthy
- `ConnectionError` from a closed session is now logged at `DEBUG` instead of `WARNING` in `poll_scheduler.py` and `keepalive.py`, eliminating log spam from in-flight requests draining out post-reconnect

Verified over 2+ hours of live runtime across 5 natural session deaths (~30 min apart), all recovering cleanly with **zero** false HA unavailability and **zero** log spam.

QuiteYellow:
Hi @aminorjourney 

Thanks for this! It's great to see this approach working on the ARTIK051 firmware and great to have it documented. A first public reference for that firmware family!

A bit of context has changed since you filed. Yesterday I merged PR #5 (production-hardening pass), which addresses the same DTLS half-open issue you hit, with a different approach: `PollScheduler.last_success_ts` as the liveness signal, `KeepaliveTask` taking a `liveness_fn` callback, and `Bridge._maybe_force_reconnect()` force-closing the session after 120s unreachable. So your bridge-level session recovery overlaps with what just landed, but not all of your changes are superseded. Yours catches a couple of corner cases PR #5 misses.

Proposed plan for merging this:

Keep:

- `samsung_appliance/appliances/fridge.py` (the whole 493-line descriptor) — no conflict, clean import.
- `samsung_appliance/appliances/__init__.py` registry add for FRIDGE.
- Your `ConnectionError → DEBUG` log-level changes in `keepalive.py` and `poll_scheduler.py`: clean noise-reduction win that PR #5 didn't cover.
- Your `ka.on_unreachable = None` nulling pattern: catches a real corner case where the dying session's keepalive can flip HA availability offline *after* the new session is already healthy. I'd port this into PR #5's `_maybe_force_reconnect` flow rather than into `publish_health`.

Drop:

- The `_session_stop` per-session Event + publish-health force-close mechanism: PR #5's `_maybe_force_reconnect` (triggered on `last_success_ts > 120s`) covers the same "session dead, restart it" goal via a different path. Running both would mean two paths force-closing the same session on the same failure, which is more risk than either alone.
- Your `bootstrap.py` change: the file itself got removed in PR #3 (it imported the pre-rename `samsung_dryer.coap` module).

Thank you for splitting into topic commits, so this maps to two commits in main: your `6c81689` (the fridge descriptor) whole, and the keepers from `0ab3d9b` (log-level DEBUG catches + `on_unreachable` nulling) as a second commit.

How does this sound? Thanks again for contributing this, and happy to talk through any of this.

aminorjourney:
This sounds great! Let's go ahead with your proposed merge - and I'll test against it on my system and report!

QuiteYellow:
Merged into main via #7, thanks again for the contribution, and looking forward to your test report!

PR #7, opened 2026-07-02
Merge PR #1: ARTIK051 fridge + session-recovery robustness

QuiteYellow:
Closes #1. Merges @aminorjourney's PR #1 by cherry-pick, per the agreed plan in the discussion thread.

## Commits

- `eca0d47` — **Fridge descriptor** (cherry-picked whole from @aminorjourney's [6c81689](https://github.com/QuiteYellow/SmartThings-Local/pull/1/commits/6c81689450cf9b91bf952f4348509cee0cdcd430), their commit + authorship preserved). Adds \`samsung_appliance/appliances/fridge.py\` (493 lines) and the FRIDGE registry entry.
- \`f81947b\` — **Session-recovery robustness** (attributed to @aminorjourney via \`git commit --author\`). Ports the keepers from their [0ab3d9b](https://github.com/QuiteYellow/SmartThings-Local/pull/1/commits/0ab3d9ba44e27bd8c6efb21916085862c7275e0f):
  - Suppress \`ConnectionError\` log noise from in-flight requests draining after a session close (poll_scheduler + keepalive at DEBUG).
  - Null \`KeepaliveTask.on_unreachable\` during a forced reconnect, so the dying session's keepalive can't flip HA availability offline after the new session is healthy. Ported into PR #5's \`_maybe_force_reconnect\` flow.
  - Drops the \`_session_stop\` / publish-health force-close mechanism from the original PR — PR #5's \`last_success_ts\` + \`_maybe_force_reconnect\` already cover the 'session dead, restart it' goal.
- \`1944324\` — **Docs**. Fridge row added to the tested-combinations table; new "Firmware families — a limitation" section under Part 1 explaining that descriptors are firmware-family-specific with no runtime feature detection; new "Fridge (ARTIK051)" section under Per-appliance notes with the capability table + firmware-specific observations; config-keys reference updated.

## Test plan

- [x] Python syntax check on all touched files.
- [x] Smoke run against live dryer + oven for 25s — bridge started cleanly, MQTT connected, both DTLS sessions established, both appliances identified by serial, seeded (25 links dryer, 16 links oven), first poll-sourced state change captured, sweep tier completed. No exceptions.
- [x] Live test against an ARTIK051 fridge — @aminorjourney to report post-merge.

### DIFF
--- a/README.md
+++ b/README.md
@@ -52,8 +52,17 @@ Read the result:
 |---|---|---|
 | Dryer | DV5000T (`DA_WM_TP2_20_COMMON`, `mnid=0AJT`) | All entities, ≤1s hot-tier poll (OBSERVE accelerates when online) |
 | Oven | NV7000BS-class (`TP1X_DA-KS-OVEN-0107X`, `mnid=0AJT`) | All entities; hot-tier poll covers door + operational state regardless of cloud reachability |
+| Fridge | ARTIK051_REF_17K (`DA-REF-ART-COMMON-1_20201124`) | Contributed by [@aminorjourney](https://github.com/aminorjourney) (PR #1). Older firmware family; port 49155, minimal `/oic/res` with full tree under `/device/0` |
 
 Other appliances on the same firmware family (washers, dishwashers, AC units) almost certainly speak the same protocol — the auth path and read primitives are common. You'd write one new descriptor in `samsung_appliance/appliances/`.
+
+### Firmware families — a limitation
+
+Descriptors are firmware-family-specific. Each file in `samsung_appliance/appliances/` hardcodes the resource layout of one firmware family: which hrefs it polls, which fields it reads, which write surfaces it exposes. There's no runtime feature detection.
+
+**What this means in practice:** if you set `APPLIANCE_<n>_CLASS=fridge` on a fridge that speaks a different firmware family than the one this descriptor was built for, the bridge will start and connect fine, but many sensors will publish as unknown and some controls won't work. Nothing catastrophic — you just get a half-broken HA device card.
+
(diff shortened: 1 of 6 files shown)


aminorjourney:
Tested on live hardware — Samsung ARTIK051_REF_17K  — 5.3+ hrs continuous runtime
Ran the merged _maybe_force_reconnect reconnect logic against a real fridge in production. Log attached (serial anonymized).

Results:

12 full DTLS session deaths, each preceded by a periodic OBSERVE refresh: DTLS session closed event (~every 30 min) that cascaded into a total session death — all polls timing out, ping failures, liveness warnings.

12/12 clean recoveries. Forced reconnect triggered consistently at 121–157s unreachable, reconnected in 0.7–1.3s, all 9 paths resubscribed, all 32 links reseeded every time.

No flapping in the offline signal — push_active → offline fires exactly once per death, no repeated thrashing.

No manual intervention required at any point.

4 of the 16 OBSERVE-refresh events did not cascade into a full death — session apparently survived the refresh on its own. (One of those four is the last event in the log window, so its outcome is unconfirmed rather than a clean pass.)

Net: reconnect mechanism holds up under real-world load, including load I didn't anticipate — didn't realize the OBSERVE refresh cycle was killing the session this often until I watched a long window. Worth a follow-up look at whether the refresh path itself can re-handshake proactively rather than always falling through to the ~130s-timeout detection, since it's firing roughly twice an hour rather than being a rare edge case. Not blocking — just flagging as a possible future optimization.

[dtls_reconnect_test.log](https://github.com/user-attachments/files/29644531/dtls_reconnect_test.log)

QuiteYellow:
Thank you for the confirmation logs! Happy to see the recovery path is holding up. 

The recurring `OBSERVE refresh: DTLS session closed` line looks like it may be a symptom rather than the cause, polling stays healthy for a few minutes after each one, and there's a thread-cleanup path on reconnect that could explain why it shows up far more often than the 6 h refresh interval would suggest. I've written it up here so it doesn't get lost: **#9**. It'll need to be sequenced with the reorg in #8 since that code is moving.

Still though, the reconnect mechanism clearly works.

Issue #6, opened 2026-06-30
Home Assistant Component


mbillow:
Howdy @QuiteYellow 😄 

First off, this is amazing work and some really cool discoveries you've made. I've spent the last couple of days playing with it and turning it into an actual Home Assistant component. I did a significant amount of refactoring to how device capabilities are discovered / registered, but also wrote all the glue to actually get this directly running in Home Assistant. 

https://github.com/mbillow/localthings

I don't just want to dump a massive PR to this from my fork, so I wanted to give you the opportunity to check out what I had made and we can chat to see what, if anything, you want to pull upstream. 

Thanks again for this! I have been meaning to do some work to understand how these devices communicate and this has saved me so much time. Really excited about moving this forward and building a fully local component for these devices. My branch adds refrigerator and dishwasher support because those are what I have, FYI.

QuiteYellow:
Hi @mbillow 

Thank you! Your implementation is looking great so far. The capability-composition model is a much better architecture than the descriptor pattern I have. I'll properly test out your implementation over the weekend. 

My goal was very much to get this info out there and spark some inspiration, and it's done exactly that! Your ha-chargepoint pattern (custom component + underlying protocol library in separate repos) fits this project exactly. If you're happy to cross-link both ways, I'll keep SmartThings-Local open as the protocol research + MQTT-bridge reference deployment only. This project can stay a narrower scope for future firmware findings and for anyone who wants to play around with this over MQTT, and add a top-of-README callout pointing HA users at localthings. What do you think?

I've also added an [MIT license](https://github.com/QuiteYellow/SmartThings-Local/blob/main/LICENSE) to this repo to keep things nice and clean.

Thanks for reaching out before sending a giant PR!

mbillow:
I'm more than fine if you want to keep the OCF / CoAP / DTLS stuff in this repo. I was really sad to see that Samsung's implementation was "special" enough that no current CoAP over UDP DTLS implementation out there would work for this. I tried a few different options but all had one or two major issues that made them incompatible. I have made some changes (retry/retransmit logic, rate limiting, device client cert CA validation, etc.) to the current implementation, but that is a much smaller review than merging in all the HA stuff.

I'm interested to understand where you want to draw the line for whats in-scope for this package and what is HA-specific enough to stay in my fork. In my mind there are distinct layers: `Protocol (CoAP-over-DTLS)`, `OCF`, `Capability Registry`, `Home Assistant API`. 

[...]

PR #8, opened 2026-07-04
Reorg into protocol/ + ocf/ + mqtt_demo/, port 3 DTLS reliability fixes

mbillow:
## Summary

Implements the layout proposed in #6: `protocol/` (CoAP + DTLS transport) and `ocf/` (polling/cache/keepalive) become the vendorable library surface; `mqtt_demo/` becomes the self-contained demo consuming them. The three appliance descriptors in `mqtt_demo/samples/` are frozen as reference samples — new appliance support now goes to [localthings](https://github.com/mbillow/localthings).

Commit-by-commit:
- Pure `git mv` reorg (protocol/, ocf/, mqtt_demo/) — no behavior changes
- `sensors.index_links` folded into `StateCache.index_device_tree`; `PollScheduler` loses its `sweep_index_fn` callback
- Three fixes ported from localthings' fork:
  - Validate device cert chain against the OCF root CA (`VERIFY_PEER` + `load_verify_locations`) instead of `VERIFY_NONE`
  - Bounded Block2 GET retry/retransmit + server-negotiated block-size tracking
  - Inter-request rate limiting (`pace()`), applied both within multi-block transfers and between per-path polls in `PollScheduler._do_tier` — RT-OCF firmware drops requests sent faster than its (undocumented, per-device) ceiling
- A handful of bugs the reorg/porting process surfaced along the way: a Dockerfile/`deploy.sh` gap that would've baked `.env` into the built image (both local and remote-deploy paths), a stale `ocf/state_cache.py` type reference back into the demo package (replaced with a minimal `Protocol`), and an import-isolation test hardening pass so it can't false-pass on an inherited `PYTHONPATH`

`setup_cert.py` is left untouched — localthings mints its own certs independently (in-memory, via `cryptography`) rather than calling into this script, so splitting it wasn't actually needed for vendoring and isn't part of this PR.

## Test plan
- [x] `pytest -v` — wire round-trip, state-cache fold, and import-isolation tests all pass (11/11)
- [x] `protocol/` + `ocf/` import with zero references to `mqtt_demo/` (enforced by `tests/test_import_isolation.py`, verified against a simulated leaked-`PYTHONPATH` false-pass scenario too)
- [x] `setup_cert.py --test` against two real appliances (10.0.0.129, 10.0.0.254): handshake OK, cert accepted (2.05 on `/oic/sec/acl`) on both
- [x] Live `/device/0` GET through the ported `DtlsCoapSession` (CA validation + retry + pacing) against both appliances: clean 2.05 responses, multi-block transfers (13 and 9 blocks respectively) completed without error


QuiteYellow:
This is great work, thanks for completing the reorg! I've pulled the branch and verified the changes against my own dryer and oven:

- **Reorg** (`protocol/` + `ocf/` + `mqtt_demo/`) — all good, 11/11 pass.
- **Cert validation (`VERIFY_NONE` → `VERIFY_PEER`)** — confirmed both my devices' server chains validate against the OCF root CA.
- **Block2 retry + SZX tracking** — checked for real on the oven's 13-block `/device/0`, completed cleanly.
- **`.env`-leak prevention** — the `.dockerignore` + `context: ..` + remote-tar `.dockerignore` shipping all line up. Good catch on the reorg introducing that.

**One thing worth a look before merge — the inter-request pacing.** It's leaves latency on the table. `pace()` is a blind full-interval sleep (`self._stop.wait(self._min_req_interval)`) called before every block of a Block2 transfer. `(n−1) × 200ms` of added sleep, a fixed 2.4s on my oven's 13-block `/device/0`. It recurs on every sweep-tier poll and, on the shared single session, stalls hot-tier polls for that time.

I A/B'd it end-to-end on the oven's 13-block `/device/0` (3 sweeps each, real sleeps):

```
[blind]    mean total 4074 ms   (pacing slept ~2444 ms)
[residual] mean total 2917 ms   (pacing slept ~1071 ms)
```

Per-block round-trips are ~40–200ms (device CBOR assembly, not just wire RTT), so a residual limiter, one that sleeps only the part of the interval not already consumed since the last real send, recovers about half the added sleep and takes the sweep from ~4.1s to ~2.9s (~1.2s faster). Not the full 2.4s, but a worthwhile cut per sweep (patch below).

I also tested directly whether within-transfer pacing is needed at all and it isn't. Running the Block2 `/device/0` fetch with within-transfer pacing disabled, on both devices, every block came back with zero retransmits: 13 GETs for 13 blocks on the oven (sustained ~10 req/s, *above* its ~8 req/s independent-poll ceiling) and 5/5 on the dryer. So the server doesn't apply its request-rate throttle to continuation GETs inside a single transfer. The within-transfer pacing can probably be dropped entirely, collapsing the sweep to its transfer floor (~1.4–1.9s on the oven). I've left that out of the patch below and kept it as a separate follow-up, since it's a behavioral change resting on that measurement and deserves its own reviewed PR.

Here's the patch (verified against the oven):

```diff
diff --git a/protocol/dtls_session.py b/protocol/dtls_session.py
--- a/protocol/dtls_session.py
+++ b/protocol/dtls_session.py
@@ -113,10 +113,14 @@ class DtlsCoapSession:
 
         self._stop = threading.Event()
         self._reader_thread = None
+        self._last_send_ts = 0.0
 
     def pace(self) -> None:
-        """Sleep one rate-limit interval. Uses _stop so session teardown wakes it."""
-        self._stop.wait(self._min_req_interval)
+        """Sleep only the part of the rate-limit interval not already consumed
+        since the last real send. Uses _stop so session teardown wakes it."""
+        remaining = self._min_req_interval - (time.monotonic() - self._last_send_ts)
+        if remaining > 0:
+            self._stop.wait(remaining)
 
     # ---- lifecycle ---------------------------------------------------
 
@@ -273,6 +277,7 @@ class DtlsCoapSession:
                 raise ConnectionError("DTLS session closed")
             try:
                 self.conn.send(datagram)
+                self._last_send_ts = time.monotonic()
                 while True:
                     o = self.conn.bio_read(65535)
                     if not o:
```

## Next steps

1. **This PR** — land the residual `pace()` patch above and merge. Everything else (reorg, cert, Block2, env) is verified on my hardware; the residual limiter is a pure improvement with no behavioral bet (~1.2s/sweep).

---

### Follow-up PR:

2. **Drop within-transfer pacing.** Remove the `if num > 0: self.pace()` in `get()`'s Block2 loop; keep pacing only between independent polls in `PollScheduler` (that's the one place the throttle actually bites). Justified by the throttle test above, continuation isn't rate-limited on either device, with the new retry path as the safety net for a rare drop. This should take the sweep to its transfer floor.

3. **Per-descriptor rate.** Replace the `_DEFAULT_RATE_LIMIT_RPS = 5.0` floor with each descriptor's measured ceiling (oven ~8 → 125ms, dryer ~14 → 71ms) so the inter-poll spacing isn't over-conservative either.

---

For this PR, only 1. is in scope; 2., 3. are follow-ups I'm happy to help with or pick up.

mbillow:
Pushed your residual-pace patch. Verified it's still correct against both my appliances: `/device/0` completes cleanly on the fridge and the dishwasher.

Also found and fixed a small regression from my own port: the Block2 retry/timeout debug logging that exists in localthings' `coap_dtls.py` got dropped when I split it into `protocol/dtls_session.py`. Restored both log lines, useful for exactly the kind of investigation below.

On the "drop within-transfer pacing entirely" follow-up, I tried to reproduce your measurement against my dishwasher (10.0.0.129), since that's the appliance where I originally saw Block2 continuation retries and added the fix. Flagging what I found because it's messier than a clean yes/no.

My first pass (sequential 10-sweep blocks, pacing on vs off) gave inconsistent results run to run. Failure rate tracked elapsed time more than the pacing toggle. So I ran a proper interleaved test instead: pacing on/off alternated every sweep within one continuous session (40 sweeps, same device-state drift for both arms). Result: 2/60 hrefs failed with pacing on vs 5/60 with it off. 

Looks like a win for keeping it:

| |     failures      | mean sweep time |
|-|-------------|------------------|
| pacing **on**  | 2/60 hrefs (3.3%) | 5.85s           |
| pacing **off** | 5/60 hrefs (8.3%) | 7.00s           |

But it isn't significant (Fisher's exact, p≈0.44), and looking closer, the failures aren't 60 independent trials. They're 3 clustered congestion events on `/device/0` (sweeps 3-5, 19, 33-35). In two of those three clusters, the paced sweep failed too, right in the middle of unpaced failures on either side. Within-transfer pacing didn't rescue the device during actual congestion. I can't honestly say I've shown it helps.

Net: I don't have solid evidence either way on whether within-transfer pacing matters for this device, but the dishwasher is clearly marginal regardless: 7 hard `/device/0` failures in 40 sweeps independent of pacing. Given that, I'd rather not drop within-transfer pacing globally on a "should be safe" basis; it's about 2s/sweep of insurance on a device I can't fully characterize yet. I'd rather go straight to your item 3 (per-descriptor rate ceilings) and let within-transfer pacing become conditional on the measured ceiling per device, rather than removing it outright in a follow-up PR now.

QuiteYellow (approved):
Residual-`pace()` patch applied verbatim and the restored Block2 retry logging both verified correct against the appliance pairs. Everything in scope for this PR is done. Approving and merging.

QuiteYellow:
Your reasoning holds and I agree, thanks for running the test.

So I am dropping item 2. Let us go straight to per-descriptor ceilings and make within-transfer pacing conditional on the measured ceiling instead of unconditional. Tracked in #10.

The residual-`pace()` patch and the restored Block2 retry logging both look correct and verified on both device pairs. Merging this now, everything in scope is done. 🚀

Thank you!

PR #11, opened 2026-07-06
Package as smartthings-local: nest protocol/ + ocf/ under one namespace, add PyPI publishing

QuiteYellow:
Follow-up to #8. Nests the two library packages under a single import
namespace and adds the PyPI packaging so `smartthings-local` can be
`pip install`ed (localthings then drops its vendored copy).

## Changes
- **Nest (history preserved):** `protocol/` → `smartthings_local/protocol/`,
  `ocf/` → `smartthings_local/ocf/` via `git mv`.
- **Imports rewritten** `protocol.*` → `smartthings_local.protocol.*` and
  `ocf.*` → `smartthings_local.ocf.*` across the ocf modules, `mqtt_demo/`
  (bridge, descriptor, samples), and tests.
- **`pyproject.toml`** — dist name `smartthings-local`, hatch-vcs versioning
  from `v*` tags, wheel ships only `smartthings_local/`.
- **`.github/workflows/publish.yml`** — build + PyPI Trusted Publishing
  (OIDC, no stored token) on `v*` tags.
- **`.gitignore`** — build artifacts (`_version.py`, `dist/`, `*.egg-info/`).
- **Demo paths** — `Dockerfile` COPY and `deploy.sh` tar allowlist collapsed
  to the single `smartthings_local/` package.

## One fix beyond the plan
`protocol/ocf_root_ca.pem` was being **dropped from the wheel**: it's tracked
but matches `.gitignore`'s `*.pem`, so hatchling's VCS file selection excluded
it — and `protocol/dtls_session.py` loads it at runtime
(`load_verify_locations`), so a pip install would fail every DTLS handshake.
Fixed with a `[tool.hatch.build] artifacts` force-include.

## Validation
- `pytest tests/` → **11 passed**
- `python -m build` → sdist + wheel, version from tag, **pem bundled in both**
- fresh `pip install` (neutral cwd) → all nested imports resolve, pem readable
  from `site-packages`, `importlib.metadata.version("smartthings-local")` works

## Release prerequisites (already done)
- PyPI pending publisher configured: `QuiteYellow/SmartThings-Local`,
  workflow `publish.yml`, environment `pypi`.

## Note for localthings
After this lands, localthings drops the vendored copy, `pip install
smartthings-local`, and imports `from smartthings_local.protocol …` /
`from smartthings_local.ocf …`.

PR #12, opened 2026-07-06
Support in-memory PEM cert/key alongside file paths in DtlsCoapSession

mbillow:
Small follow-up on top of #11, needed before localthings can actually take a dependency on this package.

localthings mints its client cert at runtime through its HA config flow and never writes it to disk. `DtlsCoapSession` currently only accepts `cert_path`/`key_path` (file-based, via `use_certificate_chain_file`/`use_privatekey_file`), which would force localthings to write its in-memory cert/key to a temp file on every connect just to migrate off its own vendored copy of this transport layer.

This adds an alternate `cert_pem`/`key_pem` constructor path (ported from localthings' own `_load_pem_chain` helper), so both cert sources work:

```python
DtlsCoapSession(host, port, cert_path=path, key_path=path)   # existing, unchanged
DtlsCoapSession(host, port, cert_pem=pem_str, key_pem=pem_str)  # new
```

Constructor validates that exactly one pair is given (raises `ValueError` otherwise). `cert_path`/`key_path` are now keyword-or-positional with `None` defaults rather than required positionals, but existing callers (`mqtt_demo/bridge.py`, `setup_cert.py`) already pass them as keywords, so nothing else needed to change.

## Validation
- `pytest tests/` → 15 passed (11 existing + 4 new for the cert-loading paths, including constructor validation and a self-signed-cert round trip through `_load_pem_chain`)
- Verified against both real appliances (fridge, dishwasher): file-path constructor unaffected, and the new PEM-pair constructor completes a real DTLS handshake + GET the same way.


QuiteYellow:
Merged this into `main` and both PRs landed together — retargeted #12's base from `packaging-smartthings-local` to `main` and did a merge commit, so #11's packaging commit and your cert-PEM commit both went in as one linear stack (and #11 auto-flipped to Merged).

Tagged `v0.1.0` off the merge and pushed it out — `smartthings-local` 0.1.0 is on PyPI now (wheel + sdist, root CA pem bundled and verified from `site-packages`).

Thanks for the clean PR and tests!

PR #13, opened 2026-07-23
Added tested machines

indykoning:
This package is amazing with the features it brings!

These are the machines i have tested using the https://github.com/mbillow/localthings integration.
Let me know if the information is correct (wether i should add/remove parts of the Model family etc.)

### DIFF
--- a/README.md
+++ b/README.md
@@ -88,7 +88,8 @@ Read the result:
 
 | Appliance class | Model family | Confirmed |
 |---|---|---|
-| Dryer | DV5000T (`DA_WM_TP2_20_COMMON`, `mnid=0AJT`) | All entities, ≤1s hot-tier poll (OBSERVE accelerates when online) |
+| Washer | WW11DG (`DA_WM_TP2_20_COMMON`) | All entities |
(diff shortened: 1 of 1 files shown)


QuiteYellow:
Thanks ! 🙏 This data point backs the "same `DA_WM_TP2_20_COMMON` controller ⇒ same protocol" claim, so I've de-speculated that note in the README.

I pushed a small follow-up polish on top of your change so the row matches the table's conventions:

- Credited you + noted the test path (via `mbillow/localthings`) on the washer row.
- Split the DV90T grouping. I'd verified `mnid=0AJT` on my DV5000T specifically, so rather than implying it applies to the DV90T too, I listed it as "reported same family." If you can grab the DV90T's `mnid` (it's in `/oic/d`), I'll fold it in properly.
- Same for the washer: if you have the `WW11DG`'s `mnid` handy, happy to add it so the row is as complete as the others.

Appreciate the contribution!

indykoning:
Both my washer and dryer also claims `0AJT`. Though i got it from
`/information/vs/0` `x.com.samsung.da.diagMnid`

i couldn't find `/oic/d` in the output of the HA component

QuiteYellow:
Brilliant, thanks! I've updated accordingly.

PR #14, opened 2026-07-26
stale session detection

vmvarga:
### Problem
Samsung always-on appliances (refrigerators) hold stale DTLS sessions after a
Home Assistant restart. The DTLS handshake succeeds (device responds at the TLS
layer), but the CoAP application layer remains unresponsive for 5–15 minutes
until the device's internal session expires.
This causes:
- `connect()` succeeds (handshake completes within 12s)
- First `get()` blocks for the full caller timeout (15–35s) waiting for a CoAP
  response that never arrives
- Combined blocking time approaches or exceeds HA's 60s startup deadline
- HA logs: *"Something is blocking Home Assistant from wrapping up the start up
  phase"*
The washer does not exhibit this — it cycles power/standby and clears DTLS state
naturally. The fridge runs 24/7.
### Solution
Add a lightweight CoAP-layer validation probe inside `connect()` that fires
immediately after the DTLS handshake succeeds:
1. Start the reader thread
2. Send a short-timeout GET `/device/0` (5s default)
3. If the device responds with 2.xx → session is healthy, proceed normally
4. If no CoAP response → tear down the session and raise `ConnectionError`
This surfaces stale sessions in ~17s (12s handshake + 5s probe) instead of
35–47s, keeping HA startup well within its 60s budget.
#### `DtlsCoapSession.connect(probe_timeout=None)`
New optional parameter. Default `None` preserves existing behavior (handshake
only). When set to a positive float, connect additionally:
- Starts the reader thread internally
- Sends a validation GET with the specified timeout
- Tears down and raises `ConnectionError` on probe failure
```python
# Before (still works, no behavioral change)
sess.connect()
sess.start_reader()
# After (recommended for always-on devices)
sess.connect(probe_timeout=DtlsCoapSession.PROBE_TIMEOUT_S)
sess.start_reader()  # idempotent — safe to call, no-op if already running
```

DtlsCoapSession.PROBE_TIMEOUT_S = 5.0
New class constant — the recommended probe timeout. Samsung RT-OCF devices respond to initial GETs in <500ms when healthy; 5s is generous.

DtlsCoapSession.start_reader() — now idempotent
Previously would spawn a duplicate reader thread if called twice. Now checks _reader_thread.is_alive() and returns immediately if the reader is already running. This makes it safe to call after connect(probe_timeout=...) which starts the reader internally.

Backwards Compatibility
connect() with no arguments behaves identically to before
start_reader() after a validated connect is a no-op (safe for existing callers that always call connect() → start_reader())
No changes to get(), post(), subscribe(), or close() signatures

### DIFF
--- a/mqtt_demo/bridge.py
+++ b/mqtt_demo/bridge.py
@@ -240,7 +240,7 @@ class PushBridge:
             key_path=self.shared.KEY_PATH,
             on_notification=self._on_notification,
         )
-        sess.connect()
+        sess.connect(probe_timeout=DtlsCoapSession.PROBE_TIMEOUT_S)
         self.session = sess
         self.session_started_ts = time.time()
         self.connect_count += 1
--- a/smartthings_local/protocol/dtls_session.py
+++ b/smartthings_local/protocol/dtls_session.py
@@ -100,6 +100,12 @@ class DtlsCoapSession:
         code, _    = sess.post(['mode','vs','0'], cbor)
         sess.close()
 
+    Or with stale-session detection (recommended for always-on devices):
+        sess = DtlsCoapSession(host, port, cert_pem=cert, key_pem=key)
+        sess.connect(probe_timeout=5.0)   # handshake + reader + CoAP probe
+        sess.start_reader()               # idempotent, safe to call again
+        ...
+
     Cert material comes from either a file pair (cert_path, key_path) or
     an in-memory PEM pair (cert_pem, key_pem) — exactly one pair required.
     The in-memory path exists for callers (e.g. an HA config flow) that
@@ -107,6 +113,7 @@ class DtlsCoapSession:
     """
 
     HANDSHAKE_TIMEOUT_S = 12.0
+    PROBE_TIMEOUT_S = 5.0        # CoAP-layer validation after handshake
     READER_RECV_TIMEOUT_S = 1.0  # short so stop_event propagates quickly
     MAX_BLOCKS = 32              # safety bound for Block2 fetches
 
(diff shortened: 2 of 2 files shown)


QuiteYellow:
@vmvarga, thanks for this! 

I dug into why the fridge holds a stale session, and it led to a likely root-cause fix. The stale association in theory exists because a crashed client never sends DTLS `close_notify`, and on restart we reconnect from a fresh ephemeral port, so the device sees a brand-new peer and keeps the orphan until its own timer reaps it (your 5–15 min). RFC 6347 §4.2.8 covers exactly this case: a ClientHello arriving on an *existing association's 5-tuple* means the peer rebooted, and the server must complete the new handshake and discard the old association. So: bind a **fixed UDP source port** per appliance, and every reconnect re-handshakes over the same 5-tuple — the orphan gets *evicted* at handshake time instead of waited out. If that holds on the fridge, recovery goes from "wait out the device timer" to "one handshake", and no probe/timeout is needed at all.

I bench-tested what I could on my hardware today:

- Same-5-tuple rehandshake is **accepted by RT-OCF**: crash-abandon a session (no `close_notify`, no deregs), reconnect bound to the same port, handshake completes over the orphan and reads work immediately. Verified on both oven and dryer.
- I could **not** reproduce your stall here, even crash-abandoning a session with 11 live OBSERVE registrations — a fresh session from a different port worked fine on the oven.

The fix is on `fix/stale-session-seed-timeout`, committed with you as co-author: an optional `local_port` on `DtlsCoapSession`, with the bridge binding a deterministic per-appliance port (`49700 + index`). Could you run that branch against the fridge and see whether the stale-session stall still happens at all? Restart the bridge the way that used to trigger the 5–15 min wedge and see if it just connects.

If the stall is gone, we merge that . If the fridge still wedges (or misbehaves on the same-port rehandshake), we add fast detection on top, although it might need to be higher than 5s.

vmvarga:
@QuiteYellow I tested your fix and restarted the system several times. There were no issues finding the fridge after reboot.

QuiteYellow:
Thanks again @vmvarga, this is now on `main` as `8c2108a`, committed with you as co-author. I'm closing this PR as superseded by that commit.

PR #22, opened 2026-08-02
fix(setup-cert): keep SHA-1 retry compatible with LibreSSL

Jason-Morcos:
## Problem

`setup_cert.py` retries a rejected SHA-1 signing command with an OpenSSL 3 provider configuration. Apple's `/usr/bin/openssl` is LibreSSL, not OpenSSL 3; LibreSSL can exit successfully after reading that incompatible configuration without executing `openssl x509`. The script then reaches `client.pem` with no certificate present, masking the original signing failure with a later `FileNotFoundError`.

This also makes the existing SHA-1 retry test fail on a clean macOS checkout.

## Root cause

`run_allow_sha1()` assumed every executable named `openssl` understands OpenSSL 3's provider configuration.

## Fix

- Inspect `openssl version` before selecting the retry environment.
- Apply the scoped provider override only to OpenSSL 3.
- Retry LibreSSL and older OpenSSL with a clean environment, removing any inherited `OPENSSL_CONF`.
- Add a deterministic regression test proving LibreSSL never receives the OpenSSL 3 configuration.

## Scope and compatibility

This does not change the generated key, certificate fields, chain, SHA-1 requirement, or any network behavior. Existing OpenSSL 3 behavior is preserved. No certificate or key material is added to the repository.

## Validation

- LibreSSL 3.3.6: `38 passed`
- OpenSSL 3.6.3: `38 passed`
- Focused setup-certificate tests on each CLI: `5 passed`
- Compile check and staged share-safety scan: passed

### DIFF
--- a/setup_cert.py
+++ b/setup_cert.py
@@ -216,6 +216,17 @@ def run(cmd, **kw):
 def run_allow_sha1(cmd):
     """Run an openssl command with SHA-1 signatures force-enabled, for
     distros whose crypto policy otherwise blocks SHA-1 signing."""
+    version = run(['openssl', 'version']).stdout.strip()
+    if not version.startswith('OpenSSL 3.'):
+        # The provider configuration below is specific to OpenSSL 3.
+        # LibreSSL can exit successfully without running the requested
+        # command when it is given that configuration, leaving no output
+        # certificate behind. Older OpenSSL releases do not need the
+        # provider override either, so retry them with a clean environment.
+        env = dict(os.environ)
+        env.pop('OPENSSL_CONF', None)
+        return run(cmd, env=env)
+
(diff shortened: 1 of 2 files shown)


Jason-Morcos:
@codex review

Jason-Morcos:
Adversarial re-review complete for exact head `6dc9dca339a9bec6106334cf822267d3a1fdd609`; no code change was necessary. The retry selects its environment from the actual CLI version, strips inherited `OPENSSL_CONF` for LibreSSL/older OpenSSL, and preserves the OpenSSL 3 override path.

Validation remains 38 passed with LibreSSL 3.3.6 and OpenSSL 3.6.3, plus 5 focused signing tests on each CLI.

@codex review

PR #23, opened 2026-08-02
feat(errors): add redacted typed transport failures

Jason-Morcos:
## Problem

The library currently reports protocol failures through broad built-ins such as `ConnectionError`, `TimeoutError`, and `ValueError`. Callers cannot reliably distinguish a closed session from a handshake failure, malformed CoAP, a bounded Blockwise abort, or a probe backend failure without parsing strings.

Several of those strings also interpolate remote endpoints, resource paths, or raw OpenSSL text. That makes normal exception reporting unsuitable for secret-safe diagnostics.

## Public API added

`smartthings_local.errors` now provides a documented hierarchy rooted at `SmartThingsLocalError`:

- endpoint and probe failures
- session, timeout, and session-closed failures
- authentication and authorization failures
- malformed-message, Blockwise, and Observe failures

Each concrete error has a stable `code`, a fixed redacted message, and a deterministic `repr`. Arbitrary detail cannot be passed to the constructor. The existing session, probe, and CoAP boundaries now emit the appropriate concrete types where they already fail today.

OpenSSL handshake failures discard the raw exception context and chain only from a fixed `DTLS backend failed` marker, so raw backend text is retained in neither exception linkage nor the formatted traceback.

## Backward compatibility

Every mapped error still subclasses the built-in callers previously caught:

- `ConnectionError` for session/probe/authentication/Blockwise/Observe errors
- `TimeoutError` for bounded session timeouts
- `ValueError` for malformed CoAP
- `OSError` for endpoint errors
- `PermissionError` for authorization errors

Constructor argument validation remains `ValueError`.

## Scope

This PR does not change socket behavior, endpoint selection, retry policy, authentication policy, CoAP response handling, or Observe lifecycle. The unused taxonomy members establish names for those later contracts without implementing them here. It is independent of #21 and #22.

## Validation

- Full suite with Python 3.14 / OpenSSL 3.6.3: `81 passed`
- Focused error, CoAP, and probe suite: `45 passed`
- Wheel and sdist builds include `smartthings_local/errors.py`
- Import directly from the built wheel: passed
- Compile check, new-file Ruff checks, and staged share-safety scan: passed

The LibreSSL compatibility fix from #22 is now included in upstream `main`; this branch is rebased on that fix and the merged validation workflow from #21.

PR #21, opened 2026-08-02
ci: add validation and package smoke tests

Jason-Morcos:
## Context

Hi! I was really impressed with this work and took it as the impetus for a broader exploration - I was able to build my own local setup connected to all my (several) Samsung appliances.  That work has exercised more of the transport and lifecycle surface than the current public package, and I would like to contribute those general learnings upstream as a sequence of small, independently reviewed changes.

This first PR intentionally changes no library implementation. It establishes the validation baseline for later transport work so compatibility, packaging, and worker cleanup remain visible as the library evolves.

## What this adds

- Python 3.11 through 3.14 test jobs, plus dependency-floor and latest jobs
- a corrected `pyOpenSSL>=23.1` minimum, matching the first release that
  provides the DTLS timeout APIs already used by the package
- wheel and sdist build, content, isolated-install, and import checks
- a compatibility baseline for the API currently consumed by LocalThings while allowing trailing optional extensions
- deterministic cleanup checks for the existing OCF background workers
- a fail-closed introduced-content share-safety check with synthetic tests

## Local validation

- `51 passed` at the Python 3.11 dependency floor and on latest dependencies
  with Python 3.12, 3.13, and 3.14
- isolated wheel and sdist installs and imports passed
- distribution contents verified against tracked package/test files
- Ruff check and format checks passed
- actionlint passed for the workflow
- share-safety and manual public-diff review passed

## Scope

There are no transport, protocol, provisioning, or other runtime implementation changes in this PR; the only package metadata change corrects the existing pyOpenSSL minimum.

PR #24, opened 2026-08-02
feat(protocol): resolve and connect UDP endpoints

Jason-Morcos:
## Dependency

This is a one-commit stack on #23 because the endpoint contract uses its public, redacted `EndpointError`. The endpoint-specific commit is `d677c72`. Once #23 lands, this branch can be rebased and the PR diff will contain only that commit.

## Problem

`DtlsCoapSession` currently creates an `AF_INET` socket unconditionally and carries the unresolved `(host, port)` tuple through every `sendto`/`recvfrom` call.

That has three concrete limitations:

1. IPv6 and scoped IPv6 endpoints cannot be used.
2. A datagram from another host using the same destination port reaches the DTLS BIO because the socket is not pinned to a peer.
3. Fixed source-port binding always uses an IPv4 wildcard, so address family and bind behavior cannot be reviewed or tested independently.

Python's socket documentation defines the IPv4 and IPv6 `getaddrinfo` tuple shapes, recommends constraining type/protocol, and recommends trying returned candidates in order: https://docs.python.org/3/library/socket.html#socket.getaddrinfo

## Public API and behavior

- Add immutable `ResolvedUdpEndpoint`, preserving exact IPv4 or IPv6 `sockaddr` values including IPv6 flow/scope IDs.
- Keep endpoint `repr` redacted to address family only.
- Add ordered, de-duplicated UDP resolution helpers constrained to `SOCK_DGRAM` / `IPPROTO_UDP`.
- Add connected-socket setup that tries the next candidate after socket, bind, or connect failure.
- Preserve optional fixed `local_port` behavior with a family-correct wildcard bind.
- Add an optional `family=` argument to `DtlsCoapSession`, defaulting to `AF_UNSPEC`.
- Expose the active resolved endpoint as `session.endpoint`.
- Clear both endpoint and resolved destination state on `close()`.
- Move the session transport to connected-socket `send`/`recv`; short sends and socket failures become redacted `EndpointError` instances.

A connected datagram socket sets the peer used by subsequent `send` operations and limits received datagrams to that peer: https://man7.org/linux/man-pages/man3/connect.3p.html

## Compatibility and scope

Existing constructor calls, certificate inputs, `local_port`, request APIs, and `ConnectionError`/ `OSError` catches remain valid. Resolution happens afresh on each `connect()`, so host address changes are picked up on reconnect.

This PR does not change probing, authentication, handshake timing, retry policy, CoAP behavior, or lifecycle ownership. Probe concurrency and ambiguity remain a separate change.

## Validation

- Full stacked suite, Python 3.14 / OpenSSL 3.6.3: `101 passed`
- Endpoint-focused suite: `20 passed`
- Connected-peer loopback isolation test: passed 25 consecutive runs
- Deterministic coverage: IPv4, scoped IPv6, resolver ordering/deduplication, family filtering, bind fallback, connect failure cleanup, fixed source-port reuse, same-port/different-host isolation, redacted tracebacks, and connected DTLS `send`/`recv`
- Wheel and sdist include `smartthings_local/protocol/endpoint.py`; built-wheel import passed
- New-module Ruff checks, compile check, cached-diff check, and staged share-safety scan: passed

The LibreSSL compatibility fix from #22 and the validation workflow from #21 are now included in upstream `main`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -79,6 +79,36 @@ failure is chained for debugging, the cause is replaced with a fixed redacted
 marker; raw backend text is not copied into the public error or its formatted
 traceback.
 
+### Resolved UDP endpoints
+
+Sessions resolve a host to a first-class `ResolvedUdpEndpoint` and use a
+connected UDP socket for the DTLS transport. Connecting the datagram socket
+pins it to the exact resolved peer, so unrelated datagrams from another host
+using the same port are discarded by the operating system. IPv4, IPv6, and
+scoped IPv6 tuples are preserved without putting the address or scope in the
+endpoint's `repr`.
+
+Address family and fixed source-port behavior are explicit and optional:
+
+```python
+import socket
+
+sess = DtlsCoapSession(
+    "device.example",
+    49154,
+    cert_pem=cert_pem,
+    key_pem=key_pem,
+    family=socket.AF_INET6,
+    local_port=56830,
+)
+sess.connect()
+assert sess.endpoint.family == socket.AF_INET6
+```
+
+The resolver retains candidate order and the socket setup tries the next
+candidate after a family, bind, or connect failure. Resolution and socket
+setup failures raise the redacted `EndpointError` documented above.
+
 For a full worked integration, the higher-level `smartthings_local.ocf` layer (`StateCache`, `PollScheduler`, `KeepaliveTask`, `ObserveRefreshTask`) coordinates tiered polling and OBSERVE on top of a session. The MQTT bridge demo below wires all of it together.
 
 ### What the demo bridge gives you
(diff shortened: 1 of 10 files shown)


Jason-Morcos:
@codex review

Jason-Morcos:
Adversarial follow-up complete for exact head `a833db2d1d427fa5a3e0b196b80ffe3d040d2905`. Resolver, socket-setup, handshake send/receive, and runtime send failures now discard raw exception context; `close()` clears both `endpoint` and the resolved destination; a runtime send regression test covers the redaction boundary; and the scoped-IPv6 fixture uses the documentation range.

Validation: 86 passed on Python 3.14 / OpenSSL 3.6.3; 20 endpoint-focused tests passed; the introduced-content safety and diff checks passed.

@codex review

Issue #9, opened 2026-07-05
Worker threads might not be getting stopped on reconnect


QuiteYellow:
First, the good news this came out of: @aminorjourney ran the merged `_maybe_force_reconnect` path (PR #7) against a live ARTIK051 fridge for ~6 h, and recovery worked, 12/12 session deaths detected and cleanly reconnected (121–157 s to trigger, 0.7–1.3 s to reconnect + full reseed). So this isn't a recovery bug. It's something the log turned up along the way that looks worth a closer look.

I think we might be leaving worker threads running after a reconnect. The thing that caught my eye: `OBSERVE refresh: DTLS session closed` shows up roughly every 30 minutes (16 times over ~6 h), but `OBSERVE_REFRESH_INTERVAL_S` is set to 6 h, so a single, correctly-scoped refresh task probably shouldn't be able to emit that line that often.

Note the OBSERVE-refresh line pacing out ahead of every death, and near the end (08:36 / 08:52 / 08:59, then 09:23 / 09:30) it starts firing more than once between deaths:

```
03:53:59  WARNING  OBSERVE refresh: DTLS session closed
03:57:55  WARNING  device unreachable after 3 ping failures
03:59:57  WARNING  unreachable for 122s — forcing session reconnect
04:00:01  INFO     seeded → 32 links; sensors live
04:24:58  WARNING  OBSERVE refresh: DTLS session closed
04:28:21  WARNING  device unreachable after 3 ping failures
04:30:57  WARNING  unreachable for 156s — forcing session reconnect
04:30:59  INFO     seeded → 32 links; sensors live
   … (cadence holds ~every 30 min) …
08:29:00  WARNING  OBSERVE refresh: DTLS session closed
08:32:24  WARNING  device unreachable after 3 ping failures
08:34:58  WARNING  unreachable for 154s — forcing session reconnect
08:35:01  INFO     seeded → 32 links; sensors live
08:36:56  WARNING  OBSERVE refresh: DTLS session closed
08:52:58  WARNING  OBSERVE refresh: DTLS session closed
08:59:00  WARNING  OBSERVE refresh: DTLS session closed
09:02:31  WARNING  device unreachable after 3 ping failures
09:04:58  WARNING  unreachable for 147s — forcing session reconnect
09:05:01  INFO     seeded → 32 links; sensors live
09:23:59  WARNING  OBSERVE refresh: DTLS session closed
[...]

PR #35, opened 2026-08-09
Add cancellable DTLS connection attempts

Jason-Morcos:
## Problem

A caller can bound `connect()`, but it cannot stop an attempt that is no longer needed. Polling a `threading.Event` adds cancellation latency and requires separate socket ownership just to interrupt a blocked receive.

## Changes

- Add keyword-only `connect(cancel=...)` using a one-way `ConnectCancellation` signal.
- Back each active subscription with a socketpair; `set()` makes the wake socket readable immediately.
- Wait on the UDP socket and wake socket together with `select()`, without a polling interval or watcher thread.
- Check cancellation before and after synchronous authentication, OpenSSL, and endpoint setup phases.
- Close only the temporary connection-attempt socket and raise the existing `SessionClosedError` when cancellation wins.
- Make the completion/cancellation race explicit: cancellation observed before the wake subscription is retired wins; a later signal does not alter the established session.
- Keep the signal reusable across simultaneous attempts and ensure every per-attempt socketpair is removed and closed.

This revision deliberately removes `quiesce_for_close()`, `abort()`, the in-progress-socket field, pending-request locking/wakeup changes, and all established-session lifecycle changes. Those should be introduced only with a concrete consumer that establishes their required shape.

An ordinary `threading.Event` has no selectable file descriptor, so using it with `select()` would require either polling or a helper thread. `ConnectCancellation` exposes the same small `set()` / `is_set()` shape while providing the socket wakeup directly.

## Validation

- 198 tests with current dependencies.
- 198 tests on Python 3.11 with pyOpenSSL 23.1.
- All 1,070 LocalThings tests against this exact checkout.
- 25 repeated cancellation stress runs.
- Focused Ruff, compile, share-safety, distribution-content checks, and isolated wheel/sdist imports.

Tests cover invalid and pre-set signals, cancellation across setup phases, immediate wake from a blocked `select()`, multiple subscribers, the handshake-success race, successful non-cancelled connection state, socket cleanup, and redacted local wakeup failures.

## Stack

Depends only on #34 at `a44930f`, which is now directly on current `main` and independent of #33. The new commit for this slice is `3f0e437`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -48,6 +48,15 @@ sess.subscribe(["operational", "state", "vs", "0"],          # OBSERVE
 sess.close()
 ```
 
+`connect()` uses a 12-second monotonic DTLS handshake deadline by default. A
+caller that needs a shorter bounded attempt can pass a positive finite value
+without changing later reader timeouts. OpenSSL's DTLS timer schedules flight
+retransmissions within that same deadline:
+
+```python
+sess.connect(timeout=4.0)
+```
+
 If the cert/key are minted at runtime and never written to disk (e.g. inside
 an HA config flow), create the provider from memory instead:
 
--- a/README.md
+++ b/README.md
@@ -57,6 +57,22 @@ retransmissions within that same deadline:
 sess.connect(timeout=4.0)
 ```
 
+Connection attempts can also use a one-way cancellation signal. The signal is
+backed by a socketpair, so setting it wakes the network wait immediately while
+OpenSSL retains control of DTLS retransmission timing:
+
+```python
+from smartthings_local.protocol.dtls_session import ConnectCancellation
+
+cancel_connect = ConnectCancellation()
+# Another thread may call cancel_connect.set().
+sess.connect(timeout=8.0, cancel=cancel_connect)
+```
+
+Setting the signal stops subscribed connection attempts and closes their
+temporary UDP sockets. It does not alter an already established session or add
+new session lifecycle methods. Interrupted attempts raise `SessionClosedError`.
+
 If the cert/key are minted at runtime and never written to disk (e.g. inside
 an HA config flow), create the provider from memory instead:
 
--- a/smartthings_local/protocol/dtls_handshake.py
+++ b/smartthings_local/protocol/dtls_handshake.py
@@ -1,0 +1,72 @@
+"""Shared memory-BIO driver for bounded DTLS handshakes."""
+
+from __future__ import annotations
+
+import time
+from collections.abc import Callable
+
+from OpenSSL import SSL
+
+from .coap import split_dtls
+
+_HANDSHAKE_POLL_S = 0.5
+_MAX_DATAGRAM_SIZE = 65535
+
+
+def _drive_dtls_handshake(
+    connection,
+    sock,
+    *,
+    deadline: float,
+    retries: int | None = None,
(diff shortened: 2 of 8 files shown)


QuiteYellow:
`cancel=` is the part I want. The rest I'd like to hold, and I'll explain the reasoning.

Three new lifecycle verbs land here: `cancel=`, `quiesce_for_close()` and `abort()`. Today the session has one, `close()`. Callers of the other two:

- `mqtt_demo/bridge.py` calls `sess.close()` and nothing else.
- mbillow/localthings calls `sess.close()` from `_close_session()` and nothing else.

So `quiesce_for_close()` and `abort()` ship as public API with no consumer in either place. I'd rather the fix for #9 land first and pull in whichever verb it turns out to need. Public surface added ahead of a caller tends to fix the shape of the caller before we know what shape it wants.

On the cancellation mechanism itself. Polling a `threading.Event` every 100 ms works, but it puts a latency floor under every cancel and it needs the in-progress-socket field to stop a blocked receive. A socketpair does the same job without either:

```python
self._wake_r, self._wake_w = socket.socketpair()
...
ready, _, _ = select.select([sock, self._wake_r], [], [], remaining)
```

Cancel writes a byte, `select` returns immediately, the receive was never blocked in the first place. That's the standard shape: libcoap and iotivity-lite run a single non-blocking event loop (`coap_io_process`, `oc_main_poll`), aiocoap gets it from asyncio, and none of them have a quiesce or abort call because interrupting a wait is a socket operation rather than a lifecycle state. It should also take a good chunk out of the 302 lines this adds to `dtls_session.py`.

I know that's a design change to code you've already written, tested and stress-run 25 times, which is the most expensive kind of review comment, but I think it's for the best here.

Same rebase note as #34. `e4fa70d` touches `dtls_session.py`, the public-API contract test and the new interruption tests, with no `auth.py` dependency, so it moves onto `main` behind #34 without waiting on the certificate profile.

Jason-Morcos:
Agreed. I’ve pulled quiesce_for_close(), abort(), the in-progress-socket state, pending-request changes, and all established-session lifecycle behavior out of this PR. It now does connection-attempt cancellation only and is stacked solely on #34, with no #33 dependency.

One implementation detail changed from the original version: an ordinary threading.Event has no selectable file descriptor, so combining it with select() would still require either polling or a watcher thread. I replaced it with a one-way ConnectCancellation signal backed by socketpairs for active attempts. set() writes the wake byte, select() returns immediately, and the per-attempt sockets are retired afterward. A signal arriving after the connection has been established does not alter that session.

Happy to change anything else as you see fit!

QuiteYellow:
Merged in v0.1.7, thanks. The socketpair-backed `ConnectCancellation` is what I had in mind.

One narrow race for a follow-up, non-blocking. The "does not alter an already established session" contract has a gap at the very end of `connect()`. If `cancel.set()` lands between `_drive_dtls_handshake` returning `completed=True` and the `finally: _unsubscribe()`, then `interrupted` comes back `True` and the completed handshake is torn down with `SessionClosedError`. It's benign (a cancel that arrives exactly at completion getting honoured is defensible), but it contradicts the comment. Either gate the teardown on `not completed`, or soften the comment to say a cancel racing completion may still win.

Issue #37, opened 2026-08-11
Samsung Refrigerator (TP2X_REF_20K) will not reconnect on HA startup


charettepa:
The Samsung Refrigerator (TP2X_REF_20K) 
Can not re-connect on HA startup
It continuously attempts to connect and fails to poll
  -"poll failed after reconnect: session operation timed out"

This can go on forever unless you perform an action in the fridge
ie. change the cooler or freezer setpoint
at which point it is able to reconnect

This behavior only started a few releases ago

I also have a Dryer (DA_WM_A51_20_COMMON) and and Stove (TP1X_DA-KS-RANGE-0101X)

The Dryer has a similar issue to the fridge
but pressing the power button on the Dryer makes it instantly re-connect

The Stove has no problems other than reporting 2 issues at startup but does re-connect
  -"device at 192.168.0.x reports serial '0JBM7DCW300421D' but this entry is registered as '192.168.0.x'; keeping the registered identity"
  -"observe /mode/vs/0: cbor decode failed: premature end of stream (expected to read 5880 bytes, got 555 instead)"

 could any of this be related to the reported issue of new firmwares having OCF endpoints and additional ports

QuiteYellow:
Thanks for the three-device comparison.

In v0.1.1 I made every reconnect use the same UDP source port. When the bridge dies without a clean shutdown, the appliance keeps holding the old session, and re-handshaking from the same port is supposed to make it drop that and accept the new one. I bench-tested that on an oven and a dryer. I never tested it on a fridge, and I said so in the commit at the time. Your fridge is the first one to report back.

In v0.1.3 the library moved to connected UDP sockets. A side effect is that the client now sees network errors it used to ignore, and one of those can kill the receive loop without logging anything. From the outside that looks exactly like what you're describing: connects, then never answers, forever.

Both fit. They point at different fixes, so I'd rather narrow it than guess.

### Two things that would tell us

**1. Which half is failing.** Add this to `configuration.yaml` and restart:

```yaml
logger:
  default: warning
  logs:
    custom_components.localthings: debug
```

Then look for `DTLS connected to 192.168.0.x:...` around the time the fridge is stuck. If that line appears and the timeout follows it, the handshake is fine and the appliance is refusing to answer reads. If it never appears, the handshake itself is failing. The message you quoted covers both cases, which is why I can't tell them apart from here.

**2. Whether an older library version behaves differently.** Edit `config/custom_components/localthings/manifest.json` and change:

[...]

PR #38, opened 2026-08-14
fix(dtls): make reader-thread death visible and fail fast

QuiteYellow:
## Problem

`_reader_loop` exited silently on any socket error with a bare `return`. Nothing was logged, and `self.conn`/`self.sock` stayed set, so the session still looked open to every caller. Each later `get()`/`post()`/`ping()` passed its `conn is None` guard, sent its CON, and waited on an `Event` no thread would ever set — three attempts at four seconds each, then `SessionTimeoutError`, on every call, forever.

This started biting in v0.1.3. `d677c72` moved the session onto **connected** UDP sockets, and a connected UDP socket surfaces ICMP errors on the next `recv`. A single `ECONNREFUSED` from an appliance whose DTLS port is momentarily closed (rebooting, router blip) now kills the reader, where an unconnected socket had simply timed out and looped.

Refs #37 — this is the invisible-failure half of what charettepa is seeing. It does **not** claim to fix a device-side wedge; it makes the failure visible and fast instead of silent and slow.

## Changes

- **Advisory ICMP errnos keep the reader alive.** `ECONNREFUSED`, `EHOSTUNREACH`, `ENETUNREACH`, `EHOSTDOWN`, `ENETDOWN` log at DEBUG and `continue` — UDP delivery was never guaranteed and the next datagram usually works.
- **Real reader exits log at WARNING.** A `close()`-driven exit (`_stop` already set) still returns quietly, so normal teardown stays silent.
- **Callers fail fast.** A new `_reader_running` Event is set in `start_reader()` and cleared in the loop's `finally`. `get`/`post`/`ping`/`subscribe`/`refresh_observes` now go through `_check_live()`, which raises `SessionClosedError` immediately when the reader is gone instead of waiting out the request timeout. Callers that never start a reader (config-flow style) are unaffected — the reader check only applies once `_reader_thread` is set.

`_send_dgram` keeps its own `conn is None` check unchanged: `_dispatch_coap` calls it to auto-ACK device CONs, so it runs on the reader thread and must not raise during teardown.

## Behaviour change for downstream (mbillow/localthings)

A dead reader now raises **`SessionClosedError`**, not `SessionTimeoutError`. `SessionClosedError` subclasses `SessionError`/`ConnectionError`, not `TimeoutError`, so localthings' `_poll_once` → `_defer_reconnect_for` three-timeout tolerance is bypassed and it reconnects on the **first** occurrence rather than after ~2 minutes of a device sitting unavailable. That is the intended outcome, but it changes reconnect timing and downstream should hear it from us.

## Tests

New `tests/test_dtls_session_reader_death.py` drives `_reader_loop` with a scripted fake socket/conn:

- advisory `ECONNREFUSED` survives and the following datagram is still dispatched (one DEBUG line, no WARNING);
- a fatal `EBADF` exits with exactly one WARNING and clears `_reader_running`;
- `get()` after reader death raises `SessionClosedError` in <1s (asserts elapsed time so a regression that reintroduces the wait is caught);
- `close()` teardown logs no WARNING;
- `_check_live()` with no reader started matches the old `conn is None` guard.

Full tracked suite: 168 passed.

## Not settled here

Whether #37's fridge is wedged by the fixed source port (v0.1.1) or by something the connected socket introduced (v0.1.3). That still needs charettepa's version-pin bisect — but once this ships, the logs will say which of the two the device is actually doing.

### DIFF
--- a/smartthings_local/protocol/dtls_session.py
+++ b/smartthings_local/protocol/dtls_session.py
@@ -18,6 +18,7 @@ Reader thread owns the UDP socket. Callers issue get()/post() and block
 on a per-token Event the reader signals. OBSERVE notifications are
 delivered via the on_notification callback.
 """
+import errno
 import os
 import socket
 import threading
@@ -71,6 +72,21 @@ _BLOCK_ACK_TIMEOUT  = 4.0
 # (200 ms) is conservative enough for all tested devices; tune per device
 # once the ceiling is measured empirically.
 _DEFAULT_RATE_LIMIT_RPS = 5.0
+
+# ICMP errors a connected UDP socket surfaces on the next recv. On these
+# appliances they show up while the device is rebooting, while it holds an
+# orphaned association, or across a router blip, and the next datagram
+# usually works. UDP delivery was never guaranteed, so treat them as
+# advisory and keep reading. Unconnected sockets never see any of this,
+# which is why the reader survived them before the connected-socket change
+# in d677c72 (v0.1.3).
+_ADVISORY_ERRNOS = frozenset(
+    value for value in (
+        getattr(errno, name, None)
+        for name in ('ECONNREFUSED', 'EHOSTUNREACH', 'ENETUNREACH',
+                     'EHOSTDOWN', 'ENETDOWN')
+    ) if value is not None
+)
 
 class DtlsCoapSession:
     """Single sustained DTLS-CoAP session.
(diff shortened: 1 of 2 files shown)


QuiteYellow:
@mbillow

mbillow:
https://github.com/mbillow/localthings/pull/368

It’ll get lumped into the next LocalThings release.

PR #40, opened 2026-08-15
fix(protocol): reassemble blockwise OBSERVE notifications

### LINKED ISSUE #39: OBSERVE notifications larger than one block are delivered truncated
## Summary

`_dispatch_coap` hands an OBSERVE notification's payload straight to `on_notification` without looking at its Block2 option (`dtls_session.py:629-643`). When a notification's representation spans more than one block, only the first block reaches the callback, and the consumer decodes a partial CBOR buffer.

One-shot `get()` is not affected: it already runs a token-stable Block2 reassembly loop (`dtls_session.py:649+`). The gap is specific to the notification path.

## Symptom

The consumer sees a CBOR decode failure on a large observed resource, most often `/mode/vs/0`. Two independent reports, same path, same error shape:

- #37 (TP1X_DA-KS-RANGE-0101X): `observe /mode/vs/0: cbor decode failed: premature end of stream (expected to read 5880 bytes, got 555 instead)`
- localthings [#361](https://github.com/mbillow/localthings/issues/361): `poll cbor decode: premature end of stream (expected to read 1 bytes, got 0 instead)` on `/mode/vs/0`

The error message belongs to the consumer, but the truncation happens here.

## Why

RFC 7959 §2.6 covers blockwise notifications. When sending a 2.05 notification the server sends only the first block of the representation, and the client retrieves the rest "as if it had caused this first response by a GET request." This library never does that second step for notifications, so anything past the first block is lost.

## Design constraints (learned on these devices)

Anyone fixing this has to account for the device behaviour already documented in the `dtls_session.py` docstring and `oven-findings.md`:

- Block2 transfers must reuse one token across every block. A fresh token per block is silently dropped.
- A fresh token has to begin its transfer at block 0. Samsung keys per-transfer state on the token, so a request that opens with Block2 NUM=1 under a token the server has not seen gets no reply at all.
- The continuation cannot borrow the notification's token. RFC 7959 §3.4 is explicit that requests for additional blocks cannot use the token of the observation relationship, and in this codebase it would also make `_dispatch_coap` resolve real notifications into the transfer's pending waiter.
- OBSERVE registrations use 1-byte tokens; Samsung silently drops TKL>1 registrations. One-shot GETs use 4-byte tokens.
- The resource can change mid-transfer. ETag validation exists for exactly that race.
- Requests are paced at 5/s. A notification storm that triggers a fetch per notification could cross the firmware ceiling.

## The fix

The two constraints above leave one shape: when a notification arrives carrying Block2 with the M bit set, discard the partial and re-read the whole resource from block 0 on a fresh 4-byte one-shot token. That is an ordinary GET, which is what §2.6 asks for, and it reuses the token-stable reassembly `get()` already performs. It costs one redundant block against a transfer that would otherwise be lost entirely.

The re-read has to be ETag-guarded against mid-transfer change and rate-limited so a notification storm cannot exceed the 5/s pacing ceiling. Deliver the reassembled representation to `on_notification` only once it is complete. `get()` already does token-stable Block2 reassembly (`dtls_session.py:649+`); factor that loop into a shared helper and reuse it here. This can land independently, on current `main`, and gives the codebase the shared Block2 primitive #36 was reaching for instead of waiting on it.

Two things need care:

- The re-read runs on the reader thread's callback path, and `get()` blocks on an event only the reader thread can set. It has to be handed to a worker, not called inline.
- When the re-read fails or has not finished (device silent, ETag mismatch mid-transfer, pacing ceiling hit), drop the notification and log at debug rather than hand the first block to the callback.

QuiteYellow:
Closes #39.

## What was wrong

`_dispatch_coap` handed an OBSERVE notification's payload to `on_notification` without looking at its Block2 option. A notification carries only the first block of a large representation (RFC 7959 §2.6), so anything past the first block was lost and the consumer decoded a partial CBOR buffer. Two reports, same path: #37 and [mbillow/localthings#361](https://github.com/mbillow/localthings/issues/361), both on `/mode/vs/0`.

## Why this re-reads instead of continuing

The issue originally proposed a continuation GET at Block2 NUM=1. I corrected that on the issue before writing any of this. Two things rule it out:

- RFC 7959 §3.4 states the requests for additional blocks cannot use the token of the observation relationship. Reusing it here would also make `_dispatch_coap` resolve real notifications into the transfer's pending waiter, since `_pending` is checked ahead of `_observe_tokens`.
- Our Block2 diagnostics recorded a fresh-token request at NUM=1 timing out on RT-OCF. The server keys per-transfer state on the token, so a token it has not seen has to open its transfer at NUM=0.

So a truncated notification is withheld and the resource is re-read from block 0 on a fresh 4-byte one-shot token. That is an ordinary GET, which is what §2.6 asks for. The MQTT bridge in this repo has been doing the same thing as a consumer-side workaround, triggered off CBOR-decode failure rather than off the Block2 option, so the shape is already proven against the oven.

## How it is wired

`get()`'s reassembly loop is extracted into `_blockwise_get`, and the notification path reuses it. That is also the shared Block2 primitive #36 was reaching for.

The re-read cannot run inline. `_dispatch_coap` runs on the reader thread and the transfer waits on an event only that thread can set, so calling it there deadlocks the session. Notifications are queued to a lazily started worker instead, latest-wins per href. The worker is serialized and paces between transfers, so a notification burst cannot push the session past the 5/s ceiling. If the re-read fails for any reason the notification is dropped at debug level; the poll tiers carry freshness, which is what they already did.

## Also fixed in the shared loop

Five things the loop got wrong, all reachable from ordinary polling and not only from the new path:

- The response's Block2 NUM was never compared against the requested one, so a retransmitted block was concatenated as if it were the next one. This is a candidate for the `expected to read 1 bytes, got 0` shape in localthings#361.
- No ETag comparison. RFC 7959 §2.4 requires it when the server supplies ETags. The transfer now restarts once on a mismatch, then raises `BlockwiseError`. None of the tested appliances emit option 4, so this is inert on them.
- A mid-transfer block-size downshift kept incrementing the block number, which then indexed the wrong byte offset. The next number now comes off the accumulated length.
- A block wait could burn its full timeout after the reader died. It now re-checks liveness while waiting and raises `SessionClosedError`, extending the #37 fail-fast contract past the first block.
- The token counters and `_pending` were mutated without a lock. The refetch worker makes the session its own second concurrent reader, so that is now guarded.

## Behaviour change for consumers

`on_notification` keeps its signature and its contract, with one difference: a truncated notification no longer reaches it at all, and the reassembled one arrives on the worker thread rather than the reader thread. Single-block notifications are unaffected and still deliver inline.

## Validation

`python -m pytest -q` on Python 3.14: 236 passed, 13 of them new in `tests/test_observe_block2.py`, which drives the reader through a loopback transport and covers inline delivery, refetch and reassembly, the fresh-token requirement, silent and error responses, burst coalescing, worker shutdown on both `close()` and reader death, and the four loop fixes. Ruff on the changed files reports nothing the file did not already report on `main`.

**Not yet validated against hardware.** The token shape the re-read uses is the assumption everything rests on, and it has only been observed indirectly through the bridge's existing fetchback. Do not tag a release off this until that is checked on a live appliance.

### DIFF
--- a/smartthings_local/protocol/coap.py
+++ b/smartthings_local/protocol/coap.py
@@ -13,6 +13,7 @@ from ..errors import MalformedMessageError
 URI_PATH       = 11
 URI_QUERY      = 15
 OBSERVE        =  6
+ETAG           =  4
 CONTENT_FORMAT = 12
 ACCEPT         = 17
 BLOCK2         = 23
@@ -119,6 +120,14 @@ def block_value(num, more, szx):
     if v <= 0xFF:    return bytes([v])
     if v <= 0xFFFF:  return struct.pack('>H', v)
     return struct.pack('>I', v)[1:]
+
+
+def block_fields(value):
+    """Decode a CoAP Block-N option value. Inverse of block_value().
+    Returns (num, more, szx). An empty value means block 0, no more,
+    SZX=0 — RFC 7959 §2.2 allows a zero-length option to elide it."""
+    v = int.from_bytes(value, 'big')
+    return v >> 4, (v >> 3) & 1, v & 0x07
 
 
 def fmt_code(c):
--- a/smartthings_local/protocol/dtls_session.py
+++ b/smartthings_local/protocol/dtls_session.py
@@ -12,11 +12,18 @@ Wire-level details that matter (from local-tools/oven-findings.md §17):
     or interleaved one-shot / OBSERVE traffic mis-attributes.
   * Multi-block GET requires the SAME CoAP token across every block
     of the response ("token-stable Block2"). Fresh-token-per-block
-    is silently dropped by the server.
+    is silently dropped by the server, and a transfer that opens at
+    NUM>0 under a token the server has not seen gets no reply at all.
 
 Reader thread owns the UDP socket. Callers issue get()/post() and block
 on a per-token Event the reader signals. OBSERVE notifications are
 delivered via the on_notification callback.
+
+A notification carries only the first block of a large representation
+(RFC 7959 §2.6). Because a continuation cannot borrow the observation's
+token (§3.4) and this server will not continue a transfer it did not
+start, such a notification is withheld and the resource is re-read from
+block 0 on a fresh one-shot token by a worker thread. See #39.
 """
 import errno
 import math
@@ -35,11 +42,12 @@ from ..errors import (
     SessionTimeoutError,
 )
 from .coap import (
-    URI_PATH, URI_QUERY, OBSERVE, CONTENT_FORMAT, ACCEPT, BLOCK2, SIZE2,
+    URI_PATH, URI_QUERY, OBSERVE, ETAG, CONTENT_FORMAT, ACCEPT, BLOCK2, SIZE2,
     TYPE_CON, TYPE_NON, TYPE_ACK, TYPE_RST,
     METHOD_GET, METHOD_POST, CF_CBOR,
     OBSERVE_REGISTER, OBSERVE_DEREGISTER, BLOCK_SZX,
-    encode_options, parse_coap, build_coap, block_value, fmt_code,
+    encode_options, parse_coap, build_coap, block_value, block_fields,
+    fmt_code,
     split_dtls as _split_dtls,
 )
 from .auth import (
@@ -72,12 +80,33 @@ DEBUG_BRIDGE = os.environ.get('DEBUG_BRIDGE') == '1'
 _BLOCK_MAX_ATTEMPTS = 3
 _BLOCK_ACK_TIMEOUT  = 4.0
 
+# How often a block wait re-checks that the reader is still alive. Short
+# enough that a mid-transfer reader death fails fast instead of burning
+# the whole per-block timeout, long enough to stay off the CPU.
+_BLOCK_LIVENESS_POLL_S = 0.25
+
 # Inter-request pacing: minimum seconds between CoAP CON sends on one session.
 # Samsung's RT-OCF stacks drop requests when hit faster than their firmware
 # ceiling (dryer ~14 req/s, oven ~8 req/s, dishwasher unknown). 5 req/s
 # (200 ms) is conservative enough for all tested devices; tune per device
 # once the ceiling is measured empirically.
 _DEFAULT_RATE_LIMIT_RPS = 5.0
+
+# Maximum hrefs held for OBSERVE refetch at once. A notification storm
+# on more resources than this is already past what the 5/s ceiling can
+# drain, so the excess is dropped rather than queued indefinitely.
+_MAX_PENDING_REFETCH = 16
+
+# Timeout for one notification refetch. Generous relative to a poll:
+# the resource is known large (that is why it blocked) and the worker
+# is serialized, so a slow one delays only later refetches.
+_REFETCH_TIMEOUT_S = 15.0
+
+
+class _EtagChanged(Exception):
+    """Internal: the server's ETag changed partway through a Block2
+    transfer, so the blocks in hand are from two different versions."""
+
 
 # ICMP errors a connected UDP socket surfaces on the next recv. On these
 # appliances they show up while the device is rebooting, while it holds an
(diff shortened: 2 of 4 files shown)


QuiteYellow:
Validated against hardware. Two independent sessions, one after a deploy and one after a process restart:

```
12:35:25  refetch /mode/vs/0 tok=bfadc91e blocks=10 bytes=10068 ok
12:45:40  refetch /mode/vs/0 tok=d50961b3 blocks=10 bytes=10068 ok
```

Different tokens, both four bytes, so the re-read used a fresh one-shot token rather than the observation's one-byte token. The two runs reassembled byte-identically, so the block count and total are deterministic rather than dependent on how the transfer interleaved with polling.

Wire dump for the first of the two, with `DEBUG_BRIDGE=1`:

```
rx ACK code=2.05 mid=cb85 tok=bfadc91e opts=7 pl=1024
rx ACK code=2.05 mid=cb88 tok=bfadc91e opts=6 pl=1024
... seven more ...
rx ACK code=2.05 mid=cb97 tok=bfadc91e opts=6 pl=852
```

Nine blocks at 1024 bytes plus one at 852 sums to exactly 10068. The message IDs are non-contiguous, so the transfer interleaved with ordinary polling and the token correlation held throughout. No `cbor decode failed` on `/mode/vs/0`, and no fetchback warnings: the consumer-side workaround in the bridge never fired, because the library handed it a complete buffer.

Those two came from the initial OBSERVE registration response rather than a spontaneous push, because the appliance normally has no internet access and its cloud-gated notify dispatch is silent. That still exercises the same code: `subscribe()` registers the token in `_observe_tokens` and creates no pending waiter, so the initial 2.05 is dispatched through the observe branch, which is the branch a spontaneous notification takes. `_dispatch_coap` cannot distinguish them.

One consequence to record: the truncation therefore fires on every session start for every observed resource larger than one block, not only when the push channel is live. That is probably why it reproduced so consistently in #37 and localthings#361, and it means the fix matters on air-gapped installs too.

### Under a real notification burst

The appliance was then given internet access briefly, with a cook started so several observed resources changed at once. Eleven re-reads over the window: six delivered, five superseded, none failed and none dropped for a full queue.

The coalescing resolves the way it should. Every run of superseded results ends in a delivery:

```
12:58:34 superseded -> :36 superseded -> :38 superseded -> :40 superseded -> 12:58:42 ok
12:58:48 superseded                                                        -> 12:58:51 ok
```

Intermediate reassemblies are discarded because a newer notification is already queued, and the final state is the one handed to the callback. Nothing is lost at the end of a chain, which was the failure mode worth checking.

End to end, on the delivery that closed the second chain:

```
refetch /mode/vs/0 tok=d5096d56 blocks=10 bytes=10066 ok
sensor diff: {'mode': "'NoOperation' -> 'Convection'", 'lamp': "'Off' -> 'On'"}
state changed [observe] (machine=active oven=Preheat temp=36/160C mode=Convection ...)
```

Ten blocks reassembled, decoded, and applied to the cache from the observe path. Before this change that notification was a truncated CBOR buffer. Payload size moved between 10066 and 10068 bytes across the window, so these were genuine state changes rather than the same representation re-served.

Pacing held. The poll counter covers scheduled polls only, so the roughly 90 block GETs the re-reads issued in that 60-second window are extra traffic contending with it rather than part of the count. Polls in the burst window: 238 ok, 0 err, 0 timeouts, against 220 to 231 in the quiet windows either side, with a peak round trip of 1959 ms against a 1830 to 2758 ms range when idle. The added Block2 traffic cost the poll path nothing measurable.

### What hardware did not cover

Three paths remain covered by unit tests only, all of them determined by the server rather than by the client:

- The ETag guard. No tested appliance emits option 4, so it stayed inert.
- The SZX downshift. The server ran every transfer at SZX=6.
- The queue-full path. Eleven re-reads never came close to the 16-href bound.

### Incidents, and what they were

Two warnings during the session, neither traceable to this change.

The first was a single oven poll timeout on `/temperatures/vs/0`, one error in a 216-request window, with 186/0 and 230/0 in the windows either side. It landed 103 seconds after the only refetch in that period, so the extra Block2 traffic was not in flight. `/temperatures/vs/0` is a small single-item resource with no blockwise path. I have no pre-change baseline to compare the rate against, so this is unattributed rather than cleared.

The second was the other appliance losing its session for about 56 seconds after a container recreate, retrying with backoff until it recovered. That is the fixed source port orphan behaviour from #37. A later process restart, which does not recreate the container, did not reproduce it, which supports pinning it on the recreate.

### Remaining caveat

This is one firmware. Burst behaviour on a different appliance could differ, and the soak so far is short: the README notes this server sometimes closes sessions right after a Block2 GET, and no session closed during the test, but half an hour is not a long enough window to say much about that. Anyone on #37 or localthings#361 who can run the branch against their own device would be adding the coverage this cannot.

PR #42, opened 2026-08-15
fix(protocol): retain completed DTLS handshakes

Jason-Morcos:
## Summary

- retain a DTLS session once OpenSSL reports that the handshake completed, even if the monotonic deadline is reached while that call returns
- keep the deadline as the bound for setup, retransmit servicing, and network waits
- document the boundary and cover both exact-deadline and just-after-deadline completion

## Why

This is the follow-up requested in [the merge comment on #34](https://github.com/QuiteYellow/SmartThings-Local/pull/34#issuecomment-5301815991).

`_drive_dtls_handshake()` currently calls `do_handshake()`, then checks the clock again before returning success. If OpenSSL completes the handshake a hair past the deadline, `connect()` tears down the completed session and raises `SessionTimeoutError`.

The deadline still prevents another setup, retry, or network-wait iteration. This changes only the result after OpenSSL has affirmatively reported completion.

## Stack

1. This PR
2. #43 — completed-handshake vs late-cancellation follow-up

The second PR is intentionally separate so each post-merge review comment has one regression and one behavior change.

## Validation

- 237 tests on Python 3.14
- 237 tests on Python 3.11 with `cbor2==5.6.0`, `pyOpenSSL==23.1.0`, and `pytest==8.0.0`
- 1,555 tests from current LocalThings `main` against this exact head
- focused deadline/interruption tests, compile check, share-safety check, wheel/sdist content check, and isolated imports from both artifacts

### DIFF
--- a/README.md
+++ b/README.md
@@ -57,6 +57,10 @@ retransmissions within that same deadline:
 sess.connect(timeout=4.0)
 ```
 
+The deadline stops further setup, retries, and network waits. If OpenSSL
+reports that the handshake completed at the deadline boundary, the completed
+session is retained rather than torn down as a timeout.
+
 Connection attempts can also use a one-way cancellation signal. The signal is
 backed by a socketpair, so setting it wakes the network wait immediately while
 OpenSSL retains control of DTLS retransmission timing:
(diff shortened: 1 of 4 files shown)


QuiteYellow:
Thanks.

Returning `True` once `do_handshake()` reports completion is the right boundary. The deadline still gates every setup, retry, and network wait, so the bound that matters is unchanged; the only case that behaves differently now is the one where OpenSSL finished and the old code threw the result away.

CI is green across the matrix. Merging.

PR #43, opened 2026-08-15
fix(protocol): keep completed sessions on late cancel

Jason-Morcos:
## Summary

- keep an established DTLS session when cancellation is first observed while retiring the connection-attempt wake subscription
- initialize the completion state before the handshake driver so cancellation/error combinations remain well-defined
- cover the exact unsubscribe race and the cancellation-plus-backend-failure path

## Why

This is the follow-up requested in [the merge comment on #35](https://github.com/QuiteYellow/SmartThings-Local/pull/35#issuecomment-5301816072).

After `_drive_dtls_handshake()` returns `completed=True`, `cancel.set()` can race the `finally` block that unsubscribes the socketpair. `_unsubscribe()` then reports `interrupted=True`, and the current condition tears down a handshake that has already completed. Gating that teardown on `not completed` makes the implementation match the documented contract that cancellation does not alter an established session.

Cancellation during setup or a blocked network wait still wins immediately. If the handshake did not complete, a cancellation observed during unsubscribe still closes the temporary socket and raises `SessionClosedError`.

## Merge order

1. #42
2. This PR

This branch contains #42's single commit plus one additional commit for the cancellation race. Once #42 merges, this PR reduces to the second commit.

## Validation

- 238 tests on Python 3.14
- 238 tests on Python 3.11 with `cbor2==5.6.0`, `pyOpenSSL==23.1.0`, and `pytest==8.0.0`
- 1,555 tests from current LocalThings `main` against this exact head
- 25 repeated cancellation-test runs
- focused deadline/interruption tests, compile check, share-safety check, wheel/sdist content check, and isolated imports from both artifacts

### DIFF
--- a/README.md
+++ b/README.md
@@ -57,6 +57,10 @@ retransmissions within that same deadline:
 sess.connect(timeout=4.0)
 ```
 
+The deadline stops further setup, retries, and network waits. If OpenSSL
+reports that the handshake completed at the deadline boundary, the completed
+session is retained rather than torn down as a timeout.
+
 Connection attempts can also use a one-way cancellation signal. The signal is
 backed by a socketpair, so setting it wakes the network wait immediately while
 OpenSSL retains control of DTLS retransmission timing:
--- a/smartthings_local/protocol/dtls_handshake.py
+++ b/smartthings_local/protocol/dtls_handshake.py
@@ -34,14 +34,16 @@ def _drive_dtls_handshake(
     bounded only by its deadline, while the diagnostic probe retains its
     explicit retry budget.
 
-    Return ``True`` only when the handshake completes before the deadline.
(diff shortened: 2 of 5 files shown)


QuiteYellow:
Merging this after #42.

Gating the teardown on `not completed` matches what `connect()` already documents: cancellation wakes the network wait, it does not tear down a session that exists. Cancellation during setup or a blocked wait still wins, which is the behaviour I care about.

The initialisation of `completed` before the try block is worth more than the one-line condition, I think. On `main` that name is only bound on the success path, and the function stays correct purely because every failure branch raises before reaching `if not completed`. That held, but it held by accident. Better to have it defined.

CI is green. Thanks for splitting this out from #42 rather than folding both into one PR.

PR #45, opened 2026-08-15
feat(protocol): add pure OwnerPSK derivation

Jason-Morcos:
Implements the pure OwnerPSK portion agreed in #28.

After an authenticated manufacturer-certificate session, IoTivity derives the 128-bit OwnerPSK in two stages:

1. expand the 48-byte TLS master secret with the TLS 1.2 SHA-256 PRF, the “key expansion” label, and server-random followed by client-random;
2. expand that cipher-sized key block with the selected OXM label, owner UUID, and device UUID.

The new helper keeps every security-relevant input explicit and validates the exact TLS/OCF shapes before deriving anything. It accepts the eight IoTivity cipher names and matching key-block lengths used by pyOpenSSL, rejects nil UUIDs and unknown ciphers, and requires the caller to choose either the standard manufacturer-certificate label or the confirmed-manufacturer label. The documentation calls out that the label comes from the selected DOXM OXM, not the appliance model.

The PRF ordering and key-block table follow IoTivity’s [CAsslGenerateOwnerPsk implementation](https://github.com/iotivity/iotivity/blob/2ca8390cced84c54a7f8338439049cd5989b3fb4/resource/csdk/connectivity/src/adapter_util/ca_adapter_net_ssl.c#L3012-L3125). Its lower adapter uses historical resource/provisioning parameter names, but the actual [ownership-transfer caller passes owner UUID first and target-device UUID second](https://github.com/iotivity/iotivity/blob/2ca8390cced84c54a7f8338439049cd5989b3fb4/resource/csdk/security/provisioning/src/ownershiptransfermanager.c#L866-L871); the helper and a fixed reverse-order regression preserve that caller contract. The two accepted label values come from IoTivity’s [security resource strings](https://github.com/iotivity/iotivity/blob/2ca8390cced84c54a7f8338439049cd5989b3fb4/resource/csdk/security/src/srmresourcestrings.c#L202-L210).

This is the same derivation path needed after the authenticated manufacturer-certificate channel on the Samsung profiles I have working locally. It stops at deterministic key derivation: obtaining authenticated TLS state remains the caller’s job, while OCF security-resource writes, ownership-transfer policy, persistence, and cloud/account work remain outside this package.

Coverage includes fixed GCM and CCM8 vectors, explicit owner/device ordering, both OXM labels, the complete immutable cipher table, exact type/length validation, nil UUID rejection, secret-safe errors, and an exact public signature contract.

Validation:

- SmartThings-Local: 252 tests on Python 3.14
- dependency floor: 252 tests on Python 3.11 with the minimum declared versions
- current LocalThings main: 1,555 tests
- wheel and sdist install/import checks
- distribution and share-safety checks

This is one commit directly on v0.1.8. It is independent of #42, #43, #36, and #44; those changes can merge in any order.

### DIFF
--- a/README.md
+++ b/README.md
@@ -171,6 +171,32 @@ the key must be exactly 16 or 32 bytes. `PskAuth` selects only
 or persist credentials. Ownership transfer and credential discovery are
 outside this package.
 
+Code that has already completed an authenticated manufacturer-certificate
+session can derive IoTivity's 128-bit OwnerPSK from the resulting TLS state:
+
+```python
+from smartthings_local.protocol.owner_psk import derive_mfg_certificate_owner_psk
+
+owner_psk = derive_mfg_certificate_owner_psk(
+    master_secret=master_secret,
+    client_random=client_random,
+    server_random=server_random,
+    owner_uuid=owner_uuid,
+    device_uuid=device_uuid,
+    cipher_name=cipher_name,
+    oxm_label=selected_oxm_label,
+)
+```
+
+The caller must supply the exact authenticated TLS values, non-nil raw OCF
+UUIDs, negotiated cipher name, and label for the selected OXM. Use
+`STANDARD_MFG_CERTIFICATE_OXM_LABEL` for `oic.sec.doxm.mfgcert` and
+`CONFIRMED_MFG_CERTIFICATE_OXM_LABEL` for
+`x.org.iotivity.conmfgcert`; do not infer the label from the appliance model.
+The helper performs deterministic key derivation only: it does not access a
+session, discover credentials, choose an ownership method, write security
+resources, run OTM, or persist the result.
+
 ### Classified errors
 
 Runtime transport failures use the public types in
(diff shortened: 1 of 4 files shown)


QuiteYellow:
I checked this against the IoTivity source rather than against the tests, all looks good.

Two non-blocking notes.

The tests are named `test_fixed_iotivity_gcm_vector` and `test_fixed_iotivity_ccm8_vector`, but the values are generated by this implementation rather than captured from IoTivity or a device. They are right, but they could not have caught a wrong key-block length, which is the failure they are named for. A comment saying where they came from would stop the next reader trusting them for more than they cover.

The table also omits the three other suites IoTivity maps to 96: `ECDH_ANON`, `ECDHE_PSK` and `ECDHE_RSA` with AES-128-CBC-SHA256. Reasonable to leave out given the constant is scoped to manufacturer certificates, but one docstring line saying so would help.

Scope is right. Deterministic derivation with no session access, no writes, no persistence, and no caller yet. Merging.

Jason-Morcos:
Followed up in #46. The two tests are now named as synthetic regression vectors, with a short provenance note explaining that they are not captures and do not independently prove the key-block lengths. The derivation docstring also calls out the three other 96-byte IoTivity mappings that are intentionally outside the manufacturer-certificate allowlist. No behavior change.

PR #46, opened 2026-08-18
docs(protocol): clarify OwnerPSK vector scope

Jason-Morcos:
## Summary

- rename the two fixed-output tests so they are clearly synthetic regression vectors rather than captures from IoTivity or a device
- record what those values do and do not prove about the derivation
- document that IoTivity's other 96-byte ECDH_ANON, ECDHE_PSK, and ECDHE_RSA mappings are intentionally outside this helper's manufacturer-certificate allowlist

## Why

This follows up on the two non-blocking notes in [the merge comment on #45](https://github.com/QuiteYellow/SmartThings-Local/pull/45#issuecomment-5318816108). The implementation is unchanged; the test provenance and cipher scope are now explicit for the next reader.

## Validation

- SmartThings-Local: 253 tests on Python 3.14
- dependency floor: 253 tests on Python 3.11 with the minimum declared versions
- current LocalThings main: 1,625 tests
- wheel/sdist install checks, distribution, compile, and share-safety checks

### DIFF
--- a/smartthings_local/protocol/owner_psk.py
+++ b/smartthings_local/protocol/owner_psk.py
@@ -80,6 +80,8 @@ def derive_mfg_certificate_owner_psk(
     It performs no session access, network I/O, ownership writes, or storage.
     The caller must supply state from an authenticated manufacturer-certificate
     session and explicitly select the OXM label used by that transaction.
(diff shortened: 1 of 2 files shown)


QuiteYellow:
Both notes covered, thanks.

Merging.

PR #48, opened 2026-08-18
feat(protocol): validate GET and POST request options

Jason-Morcos:
## Summary

- add repeated URI-query support to `post()` while preserving the existing `get()` query contract
- add keyword-only `extra_options` to GET and POST for ordered OCF extension options
- validate path/query text, option counts, numbers, ordering, value types, and sizes before sending
- keep extension options on every token-stable Block2 GET request and every optional ordinary POST retransmission
- reserve path, query, Observe, content-format, Accept, Block1, Block2, Size1, and Size2 for the transport

## Why

Higher-level consumers need query-qualified resources and a narrow way to carry OCF extension options without constructing request datagrams themselves. The Home Assistant integration I use with local Samsung appliances needs this for OCF content-version and routing options, including across Block2 continuation requests.

This is the first PY-09 request-completeness slice. DELETE remains separate so it can reuse this validation contract, and Block1 remains transport-owned in the later stack. Response interpretation, Observe relations, ownership transactions, and device policy are unchanged.

## Current base

This branch is rebuilt directly on current `main` at `d4d3e12`. It uses the shared GET request/response path from #36 and #58, and adds options to the existing #54 POST attempt loop without replacing its pacing, liveness, MID, or retry behavior. The already-merged #51 pacing commit and its tests are no longer part of this branch.

The request-option commit is `7a531d6`.

## Stack / merge order

1. #48 (this PR): validated query and extension options
2. #49: DELETE
3. #50: bounded Block1 POST uploads

#51, #36, #54, #57, #58, #61, #62, and #64 are already on `main`. Merge #48 first.

## Validation

- 458 SmartThings-Local tests against this exact head
- bytecode compilation, diff checks, and share-safety checks

### DIFF
--- a/README.md
+++ b/README.md
@@ -48,6 +48,24 @@ sess.subscribe(["operational", "state", "vs", "0"],          # OBSERVE
 sess.close()
 ```
 
+`get()` and `post()` accept repeated URI-query strings. They also accept
+ordered `(number, bytes)` options for reviewed OCF extensions while retaining
+ownership of path, query, content-format, Accept, Observe, and blockwise
+options:
+
+```python
+code, body = sess.get(
+    ["oic", "res"],
+    query=("rt=oic.r.doxm",),
+    extra_options=((2049, b"\x08\x00"),),
+)
+```
+
+Path and query text is UTF-8 encoded, and every value is size bounded before
+anything is sent. Additional options remain arbitrary bytes, must already be
+ordered by option number, and may repeat a number when the option is
+repeatable.
+
 `connect()` uses a 12-second monotonic DTLS handshake deadline by default. A
 caller that needs a shorter bounded attempt can pass a positive finite value
 without changing later reader timeouts. OpenSSL's DTLS timer schedules flight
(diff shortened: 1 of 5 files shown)


QuiteYellow:
I have not reviewed this yet, and the three of them have sat longer than they should have. Here is where it stands.

#51 and #47 are merged. #56 now holds the order for everything left, because #36, #48, #49, #50 and #54 all collide in `dtls_session.py`, and two of them build the same MID registry from opposite ends. Your three come after #36 there, on the reasoning that additive features re-express onto a rewritten substrate more easily than a rewrite absorbs three new features. If you disagree, #56 is the place to say so.

Two practical notes. This one and #49 and #50 still carry `tests/test_request_pacing.py` from when they were stacked on #51. That merged this morning, so a rebase onto `main` should shrink all three. And your declared order, #48 then #49 then #50, is what #56 proposes, so nothing to resequence.

The part I will look at hardest when I do review it is keeping extension options on every token-stable Block2 continuation request. Token-stable Block2 is where this codebase has historically got things wrong.

QuiteYellow:
Reviewed against `7a531d6`. 458 tests pass here too, and the rebase is clean on `d4d3e12`.

PR #49, opened 2026-08-18
feat(protocol): add CoAP DELETE requests

Jason-Morcos:
## Summary

- add `METHOD_DELETE` and a public `DtlsCoapSession.delete()` request primitive
- reuse #48 validation for paths, repeated queries, and ordered extension options
- pace and recheck liveness before sending the confirmable empty-body request
- register the request in the shared token/MID exchange registry and wait through `_wait_live`
- return the same `(code, payload)` shape as GET and POST
- clean both pending indices after response, reset, timeout, or reader failure

## Scope

This is the second PY-09 slice. It adds the wire-level method only: resource selection, authorization policy, ownership transactions, response interpretation, and Block1 behavior are unchanged.

The implementation is restacked on the current shared-MID and #54 liveness substrate. A bare RST is matched by MID, a response dispatched just before reader teardown still wins, and no path returns with a stale token or MID registration.

The DELETE-only commit is `11391db`.

## Stack / merge order

1. #48: validated query and extension options
2. #49 (this PR): DELETE
3. #50: bounded Block1 POST uploads

Merge #49 after #48. It no longer depends on the old #51 branch.

## Validation

- 465 SmartThings-Local tests against this exact head
- focused DELETE encoding, pacing, RST, timeout cleanup, reader-death, response/teardown race, validation, and API-contract coverage
- bytecode compilation, diff checks, and share-safety checks

### DIFF
--- a/README.md
+++ b/README.md
@@ -48,6 +48,24 @@ sess.subscribe(["operational", "state", "vs", "0"],          # OBSERVE
 sess.close()
 ```
 
+`get()` and `post()` accept repeated URI-query strings. They also accept
+ordered `(number, bytes)` options for reviewed OCF extensions while retaining
+ownership of path, query, content-format, Accept, Observe, and blockwise
+options:
+
+```python
+code, body = sess.get(
+    ["oic", "res"],
+    query=("rt=oic.r.doxm",),
+    extra_options=((2049, b"\x08\x00"),),
+)
+```
+
+Path and query text is UTF-8 encoded, and every value is size bounded before
+anything is sent. Additional options remain arbitrary bytes, must already be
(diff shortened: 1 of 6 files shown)


QuiteYellow:
Not reviewed yet, same as #48. #56 holds the order.

QuiteYellow:
Reviewed against `11391db`. 465 tests pass here, and it merges clean on `bae8606`. #48 landing took the shared base, so there is nothing to rebase.

DELETE stays single-frame with no retransmit. That is the right default for an idempotent method: a lost request times out and the caller repeats it, without a resend racing a delete the device already ran.

Merging, then #50 rebases onto this.

PR #50, opened 2026-08-18
feat(protocol): add bounded Block1 POST uploads

Jason-Morcos:
## Summary

- keep POST bodies through 1024 bytes on the existing #54 single-frame attempt loop
- route larger bodies into Block1 before allocating an ordinary POST exchange
- upload larger bodies with one token, one shared-registry MID per logical block, byte-identical per-block retransmission, and Size1 on the first block
- honor a server-requested smaller block size by byte offset
- keep URI queries and extension options on every block
- reject incomplete, contradictory, oversized, or unbounded transfers with typed errors

## Why

This implements the Block1 path used by my working Home Assistant integration for the local washer/dryer connection flow. The transport keeps each block as its own confirmable CoAP exchange instead of relying on one oversized request being reconstructed across DTLS application-data records.

I do not currently have a shareable A/B capture of the same oversized body failing without Block1 and succeeding with Block1. QuiteYellow offered to run that comparison on a dryer and oven in the review; that remains the hardware gate before merge.

This PR adds only the transport primitive. The credential or ownership payload that a caller may send, and any decision to perform that write, remain outside this change.

## Current base and retry behavior

This branch is restacked on #49 and the merged #54/#57/#58 substrate. Bodies through 1024 bytes still use the #54 ordinary POST loop unchanged. A larger body branches before an ordinary token, MID, or pending exchange is allocated.

Each Block1 chunk registers once in the shared token/MID registry, reuses one encoded datagram across its bounded retries, stops retransmitting after an empty ACK, surfaces RST by MID, and waits through `_wait_live`. The Block1 retry count is independent of `write_max_attempts`.

The Block1-only commit is `1f5b959`.

## Stack / merge order

1. #48: validated query and extension options
2. #49: DELETE
3. #50 (this PR): bounded Block1 POST uploads

Merge #50 after #49.

## Validation

- 490 SmartThings-Local tests against this exact head
- exact and partial boundaries, size downshift, stale responses, byte-identical retransmission, response-during-pace, empty ACK, RST, reader teardown, per-attempt deadlines, builder-failure cleanup, request cap, and payload cap
- wheel and sdist content checks plus isolated install/import smoke tests
- bytecode compilation and share-safety checks

### DIFF
--- a/README.md
+++ b/README.md
@@ -48,6 +48,24 @@ sess.subscribe(["operational", "state", "vs", "0"],          # OBSERVE
 sess.close()
 ```
 
+`get()` and `post()` accept repeated URI-query strings. They also accept
+ordered `(number, bytes)` options for reviewed OCF extensions while retaining
+ownership of path, query, content-format, Accept, Observe, and blockwise
+options:
+
+```python
+code, body = sess.get(
+    ["oic", "res"],
+    query=("rt=oic.r.doxm",),
+    extra_options=((2049, b"\x08\x00"),),
+)
+```
+
+Path and query text is UTF-8 encoded, and every value is size bounded before
+anything is sent. Additional options remain arbitrary bytes, must already be
+ordered by option number, and may repeat a number when the option is
+repeatable.
+
 `connect()` uses a 12-second monotonic DTLS handshake deadline by default. A
 caller that needs a shorter bounded attempt can pass a positive finite value
 without changing later reader timeouts. OpenSSL's DTLS timer schedules flight
--- a/README.md
+++ b/README.md
@@ -69,6 +69,13 @@ repeatable.
 `delete()` uses the same path, query, extension-option, timeout, and response
 contract without sending a request payload.
 
+POST bodies through 1024 bytes retain the single-request behavior. Larger
+bodies use token-stable Block1 requests under one monotonic timeout, include
+Size1 on the first request, honor a server-requested smaller block size, and
+return only the final response. Upload bodies are limited to 512 KiB and 1024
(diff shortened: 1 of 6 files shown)


QuiteYellow:
Not reviewed yet, same as #48. #56 holds the order.

The claim I want to confirm on hardware before this lands is the motivating one: that Samsung's IoTivity server will not reconstruct an oversized CoAP message split across DTLS application-data records, so a body over one 1024-byte block needs Block1 framing with a stable token. If you have a capture of the failure without Block1, link it. I have a dryer and an oven I can try it against.

Like #49 this stays on the transport side of #28. Block1 framing is a primitive; the credential and ownership payloads that motivate it are a separate question.

QuiteYellow:
Reviewed against `1f5b959`. 490 tests pass here, and it merges clean on `main` now that #49 is in.

On the gate I set: the RT-OCF source closes it for me. Stock RT-OCF reads one datagram into a fixed `COAP_MAX_PACKET_SIZE` buffer (`COAP_MAX_HEADER_SIZE` 376 + `REST_MAX_CHUNK_SIZE` 1024 = 1400 B) and has no path that reassembles an un-blocked message across records. Its Block1 request handler is what reassembles by offset, answering 2.31 mid-transfer and 4.08 / 4.13 on an incomplete or oversized body. Those are the codes this PR drives. That is stock source, so it is corroboration and not proof, but it is enough for me to stop holding on the A/B.

The reused-token handling is the part that had to be right, and it is: a delayed response for the previous chunk is range-checked against the pending one before it can resolve it.

Merging.

PR #51, opened 2026-08-18
fix(protocol): pace CoAP request sends

### LINKED ISSUE #53: OBSERVE subscribe burst bypasses the rate limit and can wedge a device at connect
### Symptom

After a container recreate the dryer connects, seeds, and then stops answering entirely. Polls time out continuously until the unreachable watchdog forces a reconnect about 170 seconds later. From 2026-08-18:

```
19:30:01  dryer  DTLS connected — subscribing 11 paths
19:30:03  dryer  seeded → 25 links; sensors live
19:30:05  dryer  sweep complete (25 links, 928ms)
19:30:10  dryer  poll /operational/state/vs/0 timeout (cooldown 5s)
          ... continuous ...
19:31:18  dryer  liveness: no successful poll in the liveness window
19:32:51  dryer  DTLS connected — subscribing 11 paths   (forced reconnect)
```

The session is fine for roughly five seconds and then the device goes silent. The oven did the same thing on a recreate on 2026-08-07.

### Cause

`subscribe()` never paces. It calls `_send_dgram` directly at `dtls_session.py:1099`, so the only throttle on the registration burst is whatever the caller does between calls. `mqtt_demo` hardcodes one:

```python
# mqtt_demo/bridge.py:387-389
for path in self.descriptor.observe_paths:
    sess.subscribe(path)
    time.sleep(0.05)
```

50ms is 20 requests per second:

| | rate |
|---|---|
| bridge subscribe loop | 20 rps |
| `_DEFAULT_RATE_LIMIT_RPS` (`dtls_session.py:93`) | 5 rps |
| dryer firmware ceiling, measured 2026-06-03 | ~14 rps |

So eleven registrations go out at four times the library's own declared limit and above the measured ceiling for that firmware, immediately followed by the seed read. The hardcoded sleep also ignores `rate_limit_rps` entirely, so tuning the limit has no effect on this path.

### The same gap downstream

LocalThings has the same loop with no delay at all, in `observe.py:239-246`:

```python
for href in hrefs:
    segs = [s for s in href.strip("/").split("/") if s]
    try:
        session.subscribe(segs)
```

Unbounded rate. That burst runs at connect, which is when appliances in #37 fail.

Two independent callers both got this wrong, which is the real finding. Pacing is currently a caller contract enforced nowhere, and `pace()` being public is not enough to make callers use it correctly. mbillow reached the same conclusion from the write side in mbillow/localthings#384.

### Fix

#51 already does most of this. It paces the block GET including block zero, `post()`, and `delete()` inside the session rather than leaving it to callers, and it correctly leaves the reader thread's auto-ACK (`:667`), `ping()` (`:1050`), and the teardown deregister (`:464`) alone, because delaying an ACK makes the device retransmit.

`subscribe()` is the request path it does not cover, and it is the one that produced the wedge above. It needs the same treatment.

#51 currently sits fourth in a stack behind #48, #49, and #50, so the pacing commit (`75282c2`) may be worth rebasing onto `main` on its own. It is a three-hunk change and does not depend on the rest of that stack.

Interim, `mqtt_demo` should call `sess.pace()` instead of `time.sleep(0.05)` so it at least honours the configured limit.

### Not #10

#10 tunes the ceiling per descriptor and makes within-transfer Block2 pacing conditional on it. It assumes pacing is applied and asks what the value should be. This is a path that applies no pacing at any value, so the two are complementary; fixing this one makes #10's per-descriptor ceilings actually reach the subscribe burst.

Jason-Morcos:
## Summary

- pace the first Block2 GET request and every later block
- pace ordinary POST requests and OBSERVE registrations
- recheck session liveness after pacing and before a request send
- keep automatic ACKs, `ping()`, and teardown deregistration immediate

## Why

The session already spaces Block2 continuations and Block1 uploads, but the first GET, an ordinary POST, and `subscribe()` could still be sent back-to-back with earlier traffic.

The hardware report in #53 found the missing case: about eleven OBSERVE registrations were sent immediately after the DTLS handshake. The session stayed healthy for roughly five seconds, then the dryer stopped answering for nearly three minutes. Pacing `subscribe()` inside the session protects every caller from that burst instead of relying on each caller to remember it.

The Home Assistant integration I use with a locally connected Samsung washer and dryer already calls `pace()` before initial and refresh subscriptions. A regression here covers that pattern and confirms the session-owned call does not add a second wait once the interval has already elapsed.

Fixes #53.

## Stack / merge order

1. #51 (this PR) is directly on current `main` and should merge first.
2. #48 adds validated query and extension options.
3. #49 adds DELETE.
4. #50 adds bounded Block1 POST uploads.

## Validation

- 274 SmartThings-Local tests on current dependencies and the Python 3.11 dependency floor
- 1,625 LocalThings tests against this exact source tree
- deterministic eleven-registration pacing, caller-side pacing compatibility, liveness cleanup, immediate-response token ordering, and unpaced control-send regressions
- wheel and sdist content checks plus isolated install/import smoke tests
- bytecode compilation and share-safety checks

### DIFF
--- a/smartthings_local/protocol/dtls_session.py
+++ b/smartthings_local/protocol/dtls_session.py
@@ -879,8 +879,8 @@ class DtlsCoapSession:
         deadline = time.time() + timeout
         szx = BLOCK_SZX   # server may negotiate down; track per-transfer
         while True:
-            if num > 0:
-                self.pace()
+            self.pace()
+            self._check_live()
             container = self._exchange_block(
                 tok, path_segs, query, num, szx, deadline)
             if 'err' in container:
@@ -1026,6 +1026,8 @@ class DtlsCoapSession:
         with self._state_lock:
             self._pending[tok] = (ev, container)
         try:
+            self.pace()
+            self._check_live()
             self._send_dgram(datagram)
             if not ev.wait(timeout):
                 raise SessionTimeoutError()
@@ -1088,6 +1090,8 @@ class DtlsCoapSession:
 
         Returns the token used (in case the caller wants to deregister
         later)."""
+        self._check_live()
+        self.pace()
         self._check_live()
         tok = self._next_observe_tok()
         href = '/' + '/'.join(path_segs)
--- a/tests/test_request_pacing.py
+++ b/tests/test_request_pacing.py
@@ -1,0 +1,169 @@
+"""Session-owned pacing for request sends."""
+
+from __future__ import annotations
+
+from unittest.mock import Mock
+
+import pytest
+
+from smartthings_local.errors import SessionClosedError
+from smartthings_local.protocol import dtls_session
+from smartthings_local.protocol.coap import (
+    METHOD_GET,
+    METHOD_POST,
+    OBSERVE,
+    TYPE_ACK,
+    TYPE_CON,
+    build_coap,
+    parse_coap,
+)
+from smartthings_local.protocol.dtls_session import DtlsCoapSession
+
+
+class _NullAuth:
+    def configure_context(self, _context):
+        return None
+
+
+def _session():
+    session = DtlsCoapSession(
+        "device.example",
+        5684,
+        auth=_NullAuth(),
+        rate_limit_rps=1_000_000,
+    )
+    session.conn = object()
+    return session
+
+
+def test_first_get_post_and_subscribe_are_paced_before_send():
+    session = _session()
+    order = []
+    requests = []
+
+    def pace():
+        order.append("pace")
+
+    def send(datagram):
+        order.append("send")
+        request = parse_coap(datagram)
+        requests.append(request)
+        _mtype, _code, mid, token, options, _payload = request
+        if any(number == OBSERVE for number, _value in options):
+            assert session._observe_tokens[token] == "/mode/vs/0"
+            return
+        session._dispatch_coap(
+            build_coap(TYPE_ACK, 0x45, mid, token, [], b"ok")
+        )
+
+    session.pace = pace
+    session._send_dgram = send
+
+    assert session.get(["device", "0"]) == (0x45, b"ok")
+    assert session.post(["mode", "vs", "0"], b"payload") == (0x45, b"ok")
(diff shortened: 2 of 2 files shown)


Jason-Morcos:
@QuiteYellow I rebased this one off main so it doesn't have to be gated as part of the stack (and instead gated the rest of the stack on this one).

QuiteYellow:
Merged. Hardware-validated on a dryer and an oven first.

The recreate was the test I cared about. A container recreate is what wedged my oven on 7 August, and it is the same shape as #37:

```
07:33:34  dryer  DTLS connected — subscribing 11 paths
07:33:38  dryer  seeded → 25 links; sensors live
07:33:35  oven   DTLS connected — subscribing 11 paths
07:33:40  oven   seeded → 16 links; sensors live
```

Both connected, seeded and stayed up, then three clean 60s poll-windows each: 0 err, 0 ping-fail, 0 timeouts, steady-state rates matching the pre-merge baseline.

Connect to seeded went from about 2s to about 4s on the dryer. That is the eleven registrations now spaced at the rate-limit interval, which is the whole point.

The sweep cost is smaller than I predicted, and I had the reason wrong. I read "25 links" as 25 requests and expected the sweep to go from 834ms to about 5s. It does not: `poll_scheduler` logs `len(indexed)` after a single blockwise GET of `/device/0`, so the tree is one transfer and your change adds one pace interval at its head rather than one per link. Measured: dryer 834ms to 1260ms, oven 2665ms to 2690ms.

This closes #53. The registration burst there is the genuine per-request case, and `test_every_subscribe_in_registration_burst_honors_rate_limit` walks exactly it.

One thing I am leaving for a separate change: `mqtt_demo` still sleeps 50ms between subscribes at `bridge.py:389`. `pace()` absorbs that rather than adding to it, which is what `test_existing_caller_pacing_before_subscribe_does_not_wait_twice` protects, so it is redundant but harmless.

Not proven by this run: a refrigerator. #37 stays open until the reporter there confirms on a release.

PR #54, opened 2026-08-19
fix(protocol): retransmit and fail fast on the write path

mbillow:
Picks up the half of mbillow/localthings#384 that was left open after #51 claimed the pacing half.

## The asymmetry

`post()` sends its datagram exactly once:

```python
self._send_dgram(datagram)
if not ev.wait(timeout):
    raise SessionTimeoutError()
```

`get()` goes through `_exchange_block`, which retransmits each block up to `_BLOCK_MAX_ATTEMPTS` at `_BLOCK_ACK_TIMEOUT` per attempt. So one lost datagram — request or ACK — is an unrecoverable write, while a read absorbs the identical loss silently. That matches the report in #384: reads keep working, three unrelated resources (`/power/vs/0`, `/temperature/desired/0`, `/wind/direction/vs/0`) intermittently don't.

The bare `ev.wait()` is a second, separate gap: it skips the `_wait_for_block` liveness slicing `get()` uses, so a reader thread dying mid-write burns the full 8 s and reports a device timeout for what is actually a dead session.

## What this does

**Liveness (active).** `_wait_for_block` becomes `_wait_live`, and `post()` waits through it. A reader death mid-write now raises `SessionClosedError` within one liveness poll instead of at the end of the caller's timeout.

**Retransmission (off by default).** `post()` retransmits the CON up to `write_max_attempts` times inside the caller's deadline, with §4.2 backoff, pacing each retransmit, and the final attempt taking whatever budget is left.

The datagram is built once and resent verbatim. That MID reuse is the load-bearing part: a server implementing §4.5 recognises the duplicate and answers from its dedupe cache instead of re-running the write. A caller-side retry can't offer that — `post()` mints a fresh MID and token per call, so a retry from above is a genuinely new request the device has no way to dedupe.

It defaults to **1 attempt — byte-for-byte today's behaviour** — per @QuiteYellow's ordering caution on #384: retransmitting into a device that's already dropping under load turns one lost write into several, and §4.5 dedupe is unverified on RT-OCF, which doesn't reliably emit RST either. With #51 landed we can see whether writes are still being lost before turning this on, and the flag is then a one-line change rather than a new feature.

**Bare-frame matching (active).** The empty ACK ("separate response coming", §5.2.2) and RST both answer a request *without carrying a token*, so `_pending` — keyed by token — could never match either. The empty ACK was dropped on the floor and RST had no branch at all (`TYPE_RST` was imported and unused). Both now resolve through a MID registry, registered before the send so a stray frame for an unknown MID can't grow it:

- empty ACK stops retransmission and waits out the caller's budget for the separate CON
- RST surfaces as `SessionError` — a rejection, not the timeout it looked like

**`refresh_observes()` dereg sweep (active).** Paced. Unlike the teardown dereg in `close()`, which wants out quickly, this one runs against a session that has to keep working afterwards, and an unpaced OBSERVE burst is what wedges an appliance (mbillow/localthings#396).

## What this deliberately does not do

**Pacing of the request paths.** `subscribe()`, `post()`, and block zero are #51's, and a second layer at the call sites would only have to be unwound when it lands. The `time.sleep(0.05)` in the `refresh_observes` subscribe sweep is left alone for the same reason. Worth being explicit: that means **the connect-time OBSERVE burst in mbillow/localthings#396 is not fixed by this PR** — it still needs #51, then a release.

**The read path's empty-ACK handling.** Only `post()` registers a MID, so `_exchange_block` still retransmits through an empty ACK — and with a fresh MID per attempt, which defeats dedupe. That is pre-existing, and it's on the path the comment itself identifies as where RT-OCF actually uses separate responses, so it deserves its own change rather than a drive-by in a write-path PR. Happy to follow up.

## Tests

`tests/test_dtls_session_post_retry.py`, on the existing `_FakeConn`/`_FakeSock` harness. Each was confirmed to fail against the unfixed code:

- the default sends exactly once — no new load on a device nobody has measured yet
- a dropped first datagram is recovered when the flag is on
- retransmits carry the **same MID and token** as the original
- a reply landing in the retry's pace window is not resent
- a reply that beat a dying reader is still returned, not discarded as a closed session
- an empty ACK stops retransmission; an RST raises `SessionError`
- attempts stay inside the caller's deadline, including the pace before a retry
- a reader death mid-`post()` raises `SessionClosedError` fast
- `refresh_observes` paces its dereg sweep

One test-harness fix worth calling out: the stubbed `_send_dgram` didn't stamp `_last_send_ts`, so `pace()` read a zero timestamp and never slept — which silently voided the deadline test that was supposed to catch the pace-overrun bug.

Full suite: 280 passed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -77,6 +77,31 @@ Setting the signal stops subscribed connection attempts and closes their
 temporary UDP sockets. It does not alter an already established session or add
 new session lifecycle methods. Interrupted attempts raise `SessionClosedError`.
 
+Reads retransmit each Block2 request; writes send once. Where a lost write
+has been shown to be the cause rather than a device that is simply refusing
+load, `write_max_attempts` lets `post()` retransmit inside the caller's own
+timeout, backing off per RFC 7252 §4.2 and pacing every retransmit:
+
+```python
+sess = DtlsCoapSession("192.0.2.100", 49154, auth=auth, write_max_attempts=3)
+```
+
+Each attempt resends the byte-identical datagram, so a server implementing
+§4.5 can recognise the duplicate and answer from its dedupe cache instead of
+re-running the write. Retrying from the caller cannot do that — a second
+`post()` mints a fresh Message ID, which is a new request. It defaults to `1`
+(send once) because retransmitting into an appliance that is already dropping
+under load turns one lost write into several, and §4.5 dedupe is unverified on
+RT-OCF.
+
+Note that `post()`'s `timeout` bounds the whole call, rate-limit pacing
+included, rather than only the wait that follows the send. Every attempt has to
+share one budget, and a caller that asked for 8 seconds should not wait 8
+seconds plus however long the limiter withheld the request. At the default 5
+req/s that is at most 200 ms of the budget; at a hand-tuned `rate_limit_rps=1.0`
+it is a full second, so a caller pairing a low rate limit with a short timeout
+should raise the timeout to match.
+
 If the cert/key are minted at runtime and never written to disk (e.g. inside
 an HA config flow), create the provider from memory instead:
 
--- a/smartthings_local/protocol/dtls_session.py
+++ b/smartthings_local/protocol/dtls_session.py
@@ -611,7 +611,13 @@ class DtlsCoapSession:
             )
             pending = list(pending_by_id.values())
             for _ev, container in pending:
-                container.setdefault('err', SessionClosedError())
+                # Only fail an exchange that has no answer yet. A response
+                # the reader dispatched before it tore down is a real one —
+                # the request finished, the session merely died after it —
+                # and both callers check 'err' before 'code', so stamping
+                # one here discards a write the device already confirmed.
+                if 'code' not in container:
+                    container.setdefault('err', SessionClosedError())
             self._pending.clear()
             self._pending_mids.clear()
         for ev, _container in pending:
--- a/smartthings_local/protocol/dtls_session.py
+++ b/smartthings_local/protocol/dtls_session.py
@@ -1312,18 +1312,23 @@ class DtlsCoapSession:
         old token gets dropped as 'stale' — acceptable for a 6h-scale
         safety net."""
         self._check_live()
+        # Paced, unlike the teardown dereg in close(): that one wants out
+        # quickly and the session is finished either way, while this one
+        # runs against a session that has to keep working afterwards, and an
+        # unpaced OBSERVE burst is what wedges an appliance until something
+        # forces a new session (LocalThings#396). The subscribe sweep below
+        # needs nothing here — subscribe() paces its own send.
         for tok, href in list(self._observe_tokens.items()):
             segs = [s for s in href.split('/') if s]
             try:
+                self.pace()
                 self._send_observe_dereg(tok, segs)
             except Exception as e:
                 logger.warning("refresh dereg %s: %s", href, e)
         self._observe_tokens.clear()
-        time.sleep(0.1)
         for path in paths:
             try:
                 self.subscribe(list(path))
-                time.sleep(0.05)
             except Exception as e:
                 logger.warning("refresh subscribe %s: %s", path, e)
 
--- a/smartthings_local/protocol/dtls_session.py
+++ b/smartthings_local/protocol/dtls_session.py
@@ -104,6 +104,14 @@ DEBUG_BRIDGE = os.environ.get('DEBUG_BRIDGE') == '1'
 _BLOCK_MAX_ATTEMPTS = 3
 _BLOCK_ACK_TIMEOUT  = 4.0
 
+# Base per-attempt wait for a write retransmission, doubled per attempt
+# (RFC 7252 §4.2). Retransmission itself is off by default: a device that
+# is already dropping under load turns one lost write into several, and
+# MID dedupe (§4.5) is unverified on RT-OCF, which does not reliably emit
+# RST either. Enable per session via write_max_attempts once pacing has
+# been shown insufficient on real hardware (LocalThings#384).
+_WRITE_ACK_TIMEOUT = 2.0
+
 # How often a request wait re-checks that the reader is still alive. Short
 # enough that a mid-exchange reader death fails fast instead of burning
 # the whole per-attempt timeout, long enough to stay off the CPU.
@@ -250,6 +258,7 @@ class DtlsCoapSession:
                  on_notification=None, mtu=1200,
                  rate_limit_rps: float = _DEFAULT_RATE_LIMIT_RPS,
                  local_port=None, family=socket.AF_UNSPEC,
+                 write_max_attempts: int = 1,
                  auth: AuthenticationProvider | None = None):
         file_supplied = cert_path is not None or key_path is not None
         memory_supplied = cert_pem is not None or key_pem is not None
@@ -283,6 +292,7 @@ class DtlsCoapSession:
         self.on_notification = on_notification  # fn(href, payload_bytes)
         self.mtu = mtu
         self._min_req_interval = 1.0 / rate_limit_rps
+        self._write_max_attempts = max(1, int(write_max_attempts))
         # Optional fixed UDP source port. A client that dies without
         # close_notify leaves an orphaned DTLS association on the device,
         # keyed to the old 5-tuple; reconnecting from a fresh ephemeral
@@ -1153,7 +1163,17 @@ class DtlsCoapSession:
 
     def post(self, path_segs, body_cbor, timeout=8.0):
         """Single-frame POST with a CBOR-encoded body. Returns
-        (code, payload_bytes). body_cbor must already be encoded."""
+        (code, payload_bytes). body_cbor must already be encoded.
+
+        timeout bounds the whole call, pacing included, so post() returns
+        within it rather than within it plus a rate-limit interval.
+
+        Retransmits the CON up to write_max_attempts times within that
+        deadline, reusing the same Message ID: that is what lets a
+        server implementing RFC 7252 §4.5 answer a duplicate from its dedupe
+        cache instead of re-running a non-idempotent write. A caller-side
+        retry cannot offer that, since it mints a fresh MID and token.
+        Defaults to one attempt — see _WRITE_ACK_TIMEOUT."""
         self._check_live()
         tok = self._next_tok()
         opts = [(URI_PATH, s.encode()) for s in path_segs]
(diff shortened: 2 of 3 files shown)


QuiteYellow:
Reviewed, and I want this. The MID-reuse argument is the right one, off-by-default is the right call for a device nobody has measured, and the `refresh_observes` dereg pacing is a good catch that I would have missed.

It needs a rebase first, and the conflict is semantic rather than textual, so I would rather hand it back than resolve it myself.

## What changed under you

#51 merged as `b3045f5`. It paces the first send of every request, `post()` included:

```python
try:
    self.pace()
    self._check_live()
    self._send_dgram(datagram)
```

Your loop paces only retransmits (`if attempt:`), which was correct when you wrote it, because `post()` did no pacing at all. Applied on top of #51 it silently drops first-send pacing for writes. Reads keep it, writes lose it, which is the asymmetry your own PR is about, pointed the other way.

## Why I am not resolving it myself

I tried. Moving `pace()` and `_check_live()` above the loop breaks two of your tests, and both are right to break.

`test_a_reply_during_the_pace_window_is_not_resent` asserts one send. With a pace before the first attempt, your stub answers before anything goes out, the `if not ev.is_set() and not acked` guard suppresses the send, and it asserts 0 == 1.

`test_an_answer_that_beat_a_dying_reader_is_still_returned` is the one that convinced me to stop. My pre-loop `_check_live()` raised `SessionClosedError` on a write the device had already confirmed. That is exactly the bug the test exists to catch, and my resolution reintroduced it at attempt 0.

Both stubs assume the first send precedes any pace. Any resolution that paces first invalidates that assumption, so this is a question about your test structure and not a merge I should be making on your behalf.

## The question

Should `post()` pay a pace interval before its first attempt?

My answer is yes. A write is a request, that is what #51 is for, and exempting writes puts the un-limited send back on the path most likely to be hit during a storm. But it restructures your stubs, so it is yours to make.

## Release

I am cutting v0.1.9 with #51 alone rather than holding it. The reporter on #37 has a fridge that will not reconnect, the cause is the OBSERVE burst, and I have already told them the fix is written and waiting on a release. This goes in the next one.

That does mean LocalThings gets the halves separately rather than together. Given #384 is about writes and #396 is about the subscribe burst, taking the burst fix now costs you nothing you were relying on.

QuiteYellow:
I opened #56 after the review above, and it changes one thing here.

Your `_inflight_mids` and #36's `_pending_get_mids` are the same registry from opposite ends, so I do not want to merge both. #56 asks you and @atc722 to agree one structure. Hold the rebase until that settles or you risk doing it twice.

The pacing question from my review still stands and is still yours: should `post()` pay a pace interval before its first attempt? That one is independent of the registry.

QuiteYellow:
`main` answers the question I left here. #51 put `self.pace()` as the first statement in `post()`, so writes already pay an interval before their first send.

Keep that through the rebase. The work is the two stubs that assume the first send precedes any pace.

One other thing that lands on your rebase: #58 took the read-path item you had set aside. `_exchange_block` now registers once outside its attempt loop and resends one datagram, so the fresh-MID-per-attempt behaviour is gone.

mbillow:
@QuiteYellow Rebased onto main (6d9cf40). Three commits now.

**The pacing question — yes, and main already does it.** `self.pace()` is the
first statement in `post()`, so attempt 0 keeps #51's `pace → _check_live →
send` ordering exactly and retransmits pace through the same loop head. No
`if attempt:` branch anymore.

You were right that it was a test-structure question. Both stubs assumed the
first send precedes any pace; they now stub `pace()` to answer on its *second*
call, so the reply lands in the retry's window — which is what they were always
about. `test_an_answer_that_beat_a_dying_reader_is_still_returned` still catches
the bug your resolution hit: reinstating an unconditional `_check_live()` at
every attempt fails it again.

**Registry — #56 is moot for this side.** `_inflight_mids` is gone; this uses
`_pending_mids` and `_MidExchange`. The empty-ACK and RST matching are #57's,
and #58 took the read-path item I'd set aside, so what's left here is the write
attempt loop and the frames that stop it. Nothing to agree with @atc722 — I just
deleted mine.

**A bug fell out of the rebase, preexisting, not mine** (2nd commit).
`_close_pending_requests()` stamps `SessionClosedError` onto every pending
container with `setdefault()`, including one already holding the response the
reader just dispatched. Both callers read `'err'` before `'code'`, so a request
the device answered surfaces as a closed session. The window is `_reader_loop`'s
own finally: it clears `_reader_running` and drains pending in the same pass a
response may have been dispatched in. Reproduced on the write path 5/5.

Fixed by only stamping an exchange that has no answer, mirroring the guard the
RST branch already applies. It's shared, so the read path gets it too — a final
block delivered as the reader exits is an answer, not a closed session. Happy to
split this out if you'd rather take it on its own.

**Two things about `post()` that aren't unchanged from the single-send version,
both deliberate and both tested:**

- `timeout` now bounds the whole call rather than the wait after the send.
  Attempts share one budget, so it has to be armed before the first pace — and a
  caller asking for 8s shouldn't wait 8s plus however long the limiter withheld
  the request. At 5 rps that's ≤200ms of 8s; at 1 rps it's a full second, which
  is why I'd rather it be the caller's second than an extra one.
- a retransmission that fails to send is best-effort. A connected UDP socket
  reports an earlier send's queued ICMP error on the next one, and the reader
  already treats those errnos as advisory — failing the exchange there made
  turning retransmission on *less* robust than leaving it off. Attempt 0 still
  raises.

**Judgement calls, all easy to back out:**

- `_wait_for_block` → `_wait_live`, since `post()` uses it now. Touches four stub
  lines in #57/#58's tests. Say the word and I'll keep the old name.
- dropped both `time.sleep()`s in `refresh_observes`. The 50ms is #59's argument
  at the sibling call site; the 100ms is subsumed by the first `subscribe()`'s
  own pace. Slightly past what I described — separate commit.
- README gained a short `write_max_attempts` section, since a flag meant to be
  flipped after measurement should be findable.

One note for review: the empty ACK stops retransmission through *both* the send
guard and the wait branch, so no single mutation trips `test_separate_ack_stops_retransmission` — removing both does. Redundant on purpose, flagged so it doesn't read as untested.

QuiteYellow:
Three things have landed since you rebased.

#36 and #61 are on `main`. #62 is the `_wait_live` rename, pulled out of your diff and landed on its own. The docstring and the `_BLOCK_LIVENESS_POLL_S` comment are your wording verbatim, so those hunks should disappear when you rebase.

With #62 in, I test-merged your branch. Three hunks, all mechanical.

`_WRITE_ACK_TIMEOUT` lands next to a comment #62 touched, so it shows up as an add against an empty side. Keep yours. In `post()`, #36 added a comment above `self.pace()` that your attempt loop replaces. Keep yours.

The third is the one to read twice, and it is why #62 went first:

```
<<<<<<< HEAD
        deadline = time.monotonic() + per_wait
=======
        deadline = time.time() + per_wait
>>>>>>> pr54-test
```

Keep `time.monotonic()`. Three lines below, the loop reads `deadline - time.monotonic()`, so mixing the two pins `slice_s` to the poll interval and removes the per-attempt timeout: the wait then ends only when the response arrives or the reader dies, so a silent device hangs the call and the retransmission never fires. Every test stubs this function out, so nothing catches it. Before #62 that one line was buried inside a docstring-sized hunk.

#36 also reaches your new test file. It dropped `TYPE_ACK` and `TYPE_RST` from `dtls_session`'s imports when the ACK/RST handling moved into `classify_coap_response`, so `ds.TYPE_ACK` stops resolving:

```diff
-from smartthings_local.protocol.coap import parse_coap
+from smartthings_local.protocol.coap import TYPE_ACK, TYPE_RST, parse_coap
```

and the same at the two call sites. Both names are in `coap.py` on your current base as well as on `main`, so it applies whenever suits, before or after the rebase.

All four resolved that way gives 419 passing here.

## Your three judgement calls

Keeping the rename, hence #62. The `refresh_observes` sleeps: agreed, and pacing the dereg sweep is the better half of that commit. README section: keep.

## `_close_pending_requests`

Leave it in #54. The library change stands alone but its tests sit in the file your first commit creates, so splitting it means rehoming them, and the window needs the reader to die in the same pass a response was dispatched. It can ride.

One ask whenever it suits: `timeout` bounding the whole call is a public API behaviour change, so it wants a line in the release notes as well as the docstring.

mbillow:
@QuiteYellow Rebased onto main (398e76e). Three commits, 421 passing here — 405 on main plus the 16 in the new file, so the two we differ by are tests I added after your test-merge.

## The four hunks

All resolved as you called them. `_WRITE_ACK_TIMEOUT` kept, #36's comment above `self.pace()` replaced by the attempt loop, `TYPE_ACK`/`TYPE_RST` imported from `coap` in the test file, and `_wait_live` keeping `time.monotonic()`. The #62 hunks vanished from the diff exactly as intended — the only `_wait_live` line left in it is `post()`'s call site, and both test files that carried the stub rename are out of the diff entirely. Thanks for landing that separately; the trap was as advertised.

## One small correction

> Every test stubs this function out, so nothing catches it.

Not any more. I took the bad resolution deliberately and ran the new file test by test: **6 of 16 catch it.**

`test_retransmit_recovers_a_dropped_datagram` fails cleanly, `1 != 2` — the per-attempt timeout never expires, so the retransmit never fires and attempt 0 gets the answer. The other five **hang**, and that distinction is worth having in the open: `_wait_live` never returns, so `post()` never regains control to check its own deadline either. It is not that the per-attempt timeout is removed — the call never returns at all, on any budget. In CI that reads as a job timeout rather than a red assertion, so the clean failure is the one doing the useful work, and the five hangs are a symptom rather than a guard.

I did not add a defence against a future clock mix. Both lines are monotonic now, and guarding a hypothetical inside a function that is yours as of #62 seemed like the wrong instinct.

## Release notes

There is no `CHANGELOG.md` and no in-repo release notes, so I put the behaviour change in the README next to `write_max_attempts` rather than inventing a changelog convention inside a bugfix PR. Line for the next release, if that is where you want it:

```
- `DtlsCoapSession.post()`'s `timeout` now bounds the whole call, rate-limit pacing included, rather than only the wait after the send. At the default 5 req/s that is at most 200 ms of the budget; callers pairing a low `rate_limit_rps` with a short write timeout should raise the timeout to match.
```

## Everything else you raised

`_close_pending_requests` stays here as its own commit — agreed on the reasoning, its tests do sit in the file the first commit creates. Rename kept, and it is #62's now. `refresh_observes` sleeps dropped, pacing the dereg sweep left as the substance of that commit. README section kept and extended.

A note for whoever reads the diff: The empty ACK stops retransmission through both the send guard and the wait branch, so no single mutation trips `test_separate_ack_stops_retransmission` — removing both does. That redundancy is deliberate. And each of the three commits is green standalone, so this bisects cleanly.

QuiteYellow:
Merged.

The rebase is clean, all four resolutions went the way I described, and the suite is at 421 here. Both things you flagged check out. Reverting the `_close_pending_requests` guard fails both of its tests, so it earns its place. And your note about the empty ACK being covered twice is accurate.

Two follow-ups, both small, whenever suits.

## A write can go out and get no wait for its answer

Arming `deadline` before the pace means `pace()` spends the caller's budget, so when the rate-limit interval approaches the timeout there is nothing left by the time the datagram is on the wire. At `rate_limit_rps=1.0` with `timeout=1.0`:

| | write sent | wait after the send |
|---|---|---|
| before | yes | 1.00s |
| now | yes | 0.00s |

The caller is told it timed out on a write the device may have run, which is the fresh-MID caller-side retry this PR is an argument against. The default 5 req/s and 8s is nowhere near it, so the bridge is clear today. The README's own `rate_limit_rps=1.0` example is in range.

`_exchange_block` already guards this. Its deadline check at `:1080` sits immediately before `_send_dgram` and raises there, so nothing reaches the wire. Mirroring that in `post()` is three lines, though it does land on a third behaviour: the write is never attempted, where today it goes out and is simply given no time to answer. I think that is right for something non-idempotent, and it makes your README paragraph accurate about what happens. Your call though, since the deadline restructure is yours.

## The empty-ACK send guard can be tested on its own

You are right that no single mutation trips `test_separate_ack_stops_retransmission`. The guard is still reachable by itself: put the ACK in the retry's pace window and the current code sends one datagram, while dropping `if not acknowledged:` sends two. `_reply_on_second_pace` is already the right shape for it.

Thanks for the turnaround here.

PR #58, opened 2026-08-22
fix(protocol): reuse one MID across block retransmissions

QuiteYellow:
## Summary

Three follow-ups on top of #57, all named in its review. None of them are regressions from that PR.

## The read path now retransmits one MID

`_exchange_block` registered a new exchange on every attempt, so an unacknowledged retry went out under a fresh Message ID. RFC 7252 §4.2 defines a retransmission as the same message; a new ID presents each retry to the appliance as a separate request that it may answer more than once. The Block2 NUM check discarded the surplus answer, which masked the effect without removing the cause.

Registration and the datagram now happen once, outside the attempt loop, and each attempt sends identical bytes. @mbillow described this in the body of #54 and it predates #57.

Retransmission pacing is untouched, since #54 owns that policy.

## Two allocation failures have their own types

A CoAP RST raises `SessionResetError`. It refuses one exchange while the transport stays up, which is a different situation from a closed session.

Both "no free Message ID" and "this token is already in flight" raise `SessionIdentifierError`.

Previously all three surfaced as a bare `SessionError()`, which #23 added the typed hierarchy to avoid. `tests/test_dtls_session_mid_registry.py` pinned the old exact type, so that assertion moves with the behaviour.

## MID allocation stops walking the whole space

`_next_available_mid_locked` probed all 65,536 identifiers before raising. Candidates are consecutive and so distinct, which means one more candidate than there are live exchanges always finds a free one. That is a single probe in normal operation, and it only fails when every identifier is held.

## Validation

- `pytest tests/`: 325 passed, 5 of them new
- Ruff clean on the added test module and on `errors.py`
- `git diff --check` clean

## Sequencing

#56 asks everyone to hold their rebases until this is in, so that #36, #54 and #48, #49, #50 each rebase once. Merging this releases them.

### DIFF
--- a/smartthings_local/errors.py
+++ b/smartthings_local/errors.py
@@ -10,6 +10,8 @@ __all__ = [
     'ProbeError',
     'SessionClosedError',
     'SessionError',
+    'SessionIdentifierError',
+    'SessionResetError',
     'SessionTimeoutError',
     'SmartThingsLocalError',
 ]
@@ -80,6 +82,30 @@ class SessionClosedError(SessionError):
 
     code = 'session_closed'
     message = 'session is closed'
+
+
+class SessionResetError(SessionError):
+    """The peer rejected an exchange with a CoAP RST.
+
+    Distinct from a closed session: the transport is still up and every
+    other exchange on it is unaffected. Only the message the RST names has
+    been refused.
+    """
+
+    code = 'session_reset'
+    message = 'peer reset the exchange'
+
+
+class SessionIdentifierError(SessionError):
+    """No free Message ID or token was available for a new exchange.
+
+    Every identifier in the space is held by an exchange that has not yet
+    completed, which in practice means requests are leaking rather than
+    that the session is genuinely saturated.
+    """
+
+    code = 'session_identifier'
+    message = 'no free message identifier for a new exchange'
 
 
 class MalformedMessageError(SmartThingsLocalError, ValueError):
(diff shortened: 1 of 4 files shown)


QuiteYellow:
Deployed this to my two-appliance reference bridge before merging.

The happy path is unaffected. Across half an hour the dryer held 93 to 96 reads per 60s window at a p_max of 264 to 1066ms, and the oven 220 to 229 at 1887 to 2550ms, with zero errors, zero timeouts and zero ping failures on both. That matches the pre-deploy baseline window for window.

The retransmission path itself is untested on hardware. `timeouts=0` in every window means `_exchange_block` was answered on its first attempt every time, so no retry ever fired and the changed branch never executed. Forcing one needs `_BLOCK_ACK_TIMEOUT` lowered, and that constant is module-level: any value low enough to make the dryer retry also makes every oven read retry three times, which puts roughly 11.5 req/s into a device I have measured at about 8. Isolating it means taking the oven out of the bridge for the window, so I have left it for now.

So the open question from the review stands. If a Samsung stack deduplicates a repeated Message ID without a cached response to replay, a retry that used to be answered under a fresh ID would now be dropped, and a block read that recovered on attempt 2 would instead burn its whole deadline. Nothing here rules that out.

One dryer dropout during the container recreate, 44 seconds, self-recovered. Two of its reconnects returned `no live DTLS server found`, and a client-side change cannot remove the device's own listener, so that is this unit wedging on an unclean restart.

PR #65, opened 2026-08-25
feat(protocol): add bounded plaintext OCF resource reads

Jason-Morcos:
## Summary

- Add `read_plaintext_ocf_resource()` for one known absolute href and public request port.
- Reuse the source-bound GET/Block2 engine from advertised endpoint discovery instead of adding a second UDP implementation.
- Return an immutable result with the raw response code and body, Block2 metadata, attempt count, and fixed failure classification.

## Why

Some Family Hub responses span multiple Block2 packets and can arrive from a different UDP source port than the request port. Callers that need `/oic/d`, `/oic/res`, or another public resource currently have to duplicate the token, source-pinning, continuation, and deadline logic already used by discovery.

The latest hardware results in #16 also show why the response code needs to remain visible: a reachable endpoint or completed handshake does not imply that every resource is authorized. A complete 4.xx or 5.xx response is therefore returned as a complete non-success read, with its body intact.

## Behavior

- Validates the absolute href, query values, port, deadline, retries, and address family before opening a socket.
- Keeps one token across retries and Block2 continuations, uses a new MID for each NON request, and pins the first correlated response peer.
- ACKs only accepted correlated CON responses.
- Applies one deadline and the existing 32-block, 64-KiB, and datagram bounds.
- Leaves CBOR schema decoding, authentication, and authorization decisions to the caller.

This branch is directly on `main` after #48 and is independent of #49 and #50; it does not touch `dtls_session.py`.

## Validation

- 67 focused discovery/resource-read tests
- 492 full tests on Python 3.14
- 492 full tests on Python 3.11 with the dependency floor
- wheel/sdist content checks and isolated import tests
- share-safety check

Related: #16

### DIFF
--- a/README.md
+++ b/README.md
@@ -333,6 +333,45 @@ Returned ports are unauthenticated candidates, not DTLS endpoints; directory
 parsing, DTLS liveness, and authenticated device identity remain separate
 checks.
 
+### Bounded plaintext OCF resource reads
+
+Once the public request port is known, callers can read an absolute OCF href
+without reimplementing the source-port and Block2 handling used by directory
+discovery:
+
+```python
+import cbor2
+
+from smartthings_local.protocol.ocf_discovery import (
+    read_plaintext_ocf_resource,
+)
+
+resource = read_plaintext_ocf_resource(
+    "192.0.2.20",
+    "/oic/d",
+    port=5683,
+)
+if resource.successful:
+    device = cbor2.loads(resource.payload)
+elif resource.complete:
+    print(f"appliance returned CoAP code {resource.code:#04x}")
+else:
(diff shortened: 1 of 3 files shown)


QuiteYellow:
Reviewed against `552060b`. 492 tests pass here, and it merges clean on `main`.

It routes plaintext reads through the same parameterized Block2 engine discovery already uses, so the source-pinning and continuation bounds are the ones proven there. `PlaintextOcfResourceResult` keeps a complete 4.xx / 5.xx response visible, which is the distinction #16 turns on: a reachable endpoint and an authorized one are two separate things.

Merging.

Issue #75, opened 2026-08-31
Authenticated first-use identity and HVR peer cleanup


Jason-Morcos:
This follows the boundary settled in #28: these are authentication/session
mechanics only. Nothing here writes OCF security resources, selects an OTM,
or persists credentials.

Two gaps remain between the current package and the connection path I have
running against two WD53-family appliances.

## 1. Learn a verified hardware-certificate UUID during first use

`SamsungServerProfile.bound_device()` is the right runtime default: it needs a
previously verified certificate UUID and rejects every other leaf. During the
initial manufacturer-certificate flow, though, that UUID is not yet available
to the caller. The appliance presents its Samsung hardware certificate during
the first authenticated DTLS session.

The retained integration code currently does three things on that same live
session:

1. verifies the CA chain and the exact Samsung `C` / `O` / `OU` subject role;
2. extracts the non-zero UUID from the verified leaf's common name; and
3. binds that certificate UUID to the OCF device identity read from `/oic/d`
   before the association is persisted.

The certificate UUID and OCF device UUID are separate identities; the code
does not assume that they match. It also does not accept an arbitrary first
certificate. The proposed package API is an explicit first-use profile which
keeps all of the existing chain and role checks, exposes the verified
certificate UUID on the completed session, and retains no learned identity in
the reusable authentication provider. Normal reconnects continue to use
`bound_device()`.

## 2. Retire an HVR-only half-open peer

On both WD53 appliances, a failed initial handshake can stop after the DTLS
HelloVerifyRequest exchange. The appliance then retains the half-open peer on
that UDP tuple and ignores a clean retry long enough to make setup look
unrecoverable.

The retained integration handles only the narrowly proven transcript: a fixed
non-zero local UDP port, at least two ClientHello records, and inbound
handshake messages consisting only of HelloVerifyRequest. It sends one
epoch-zero fatal `handshake_failure` alert for that half-open peer and returns
control to the caller, which owns the settle delay and any retry.

I propose keeping this opt-in and fail-closed:

- require a fixed non-zero `local_port`;
- retain only bounded handshake metadata, not packet contents;
- emit the alert only when the complete observed transcript matches the
  HVR-only shape;
- never retry automatically inside `connect()`; and
- raise a distinct redacted session error so the caller can decide whether a
  retry is appropriate.

I am splitting these into two small PRs so the identity contract and cleanup
wire behavior can be reviewed independently. A final documentation/API
[...]

PR #77, opened 2026-08-31
Add opt-in HVR-only peer cleanup

Jason-Morcos:
Implements the second half of #75.

## Why

On both WD53 appliances in my local setup, an unsuccessful initial certificate
handshake can stop after HelloVerifyRequest. The appliance retains that
half-open IoTivity peer on the client UDP tuple, and an immediate clean retry
can remain silent even though the endpoint is otherwise healthy.

The working recovery path sends the standard epoch-zero fatal
`handshake_failure` alert for that exact incomplete peer, waits outside the
transport, and then retries from the same fixed source port. This PR adds the
transport half only; the package does not choose the retry delay or perform a
retry.

## What changes

- adds opt-in `connect(cleanup_hvr_peer=True)` for sessions using an actual
  `SamsungServerProfile` and a fixed non-zero `local_port`;
- records at most 32 datagrams/records as counters and one 13-byte outbound
  header, never handshake payloads;
- requires two complete epoch-zero ClientHellos, including one after an HVR;
- requires every received record to contain only complete epoch-zero
  HelloVerifyRequest messages;
- rejects malformed, fragmented, mixed, non-HVR, non-ClientHello, or oversized
  transcripts without sending an alert;
- sends at most one nonblocking cleanup alert after the handshake deadline;
  and
- raises `HandshakePeerCleanupError`, a `SessionTimeoutError` subclass, only
  when the complete alert datagram was handed to the UDP socket.

An ordinary timeout remains `SessionTimeoutError`. Cancellation, OpenSSL
failure, socket I/O failure, an inexact transcript, or cleanup-send failure do
not enter the caller's cleanup/retry path.

## Validation

- 714 tests pass on Python 3.14
- focused coverage checks the exact alert bytes and record sequence, transcript
  ordering, metadata bounds, malformed and mixed records, cancellation, send
  failure, provider/port preconditions, and no automatic retry
- the shared handshake driver reports only records successfully sent
- share-safety check passes against #76's head

## Merge order

This is stacked on #76 and currently includes that commit in the GitHub diff.
Please merge #76 first; this PR then reduces to commit `92ec459`. #74 remains
independent. The final PY-18 API/documentation PR will stack on this one.


QuiteYellow:
I read the cleanup mechanism back against the pinned RT-OCF tree I keep locally (Samsung/RT-OCF, last pushed 2023-04-07) and it holds. `rt_ssl_error_check` (`messaging/transport/rt_ssl.c:373`) frees the peer with `ssl_remove_peer_from_list` on any non-exempt negative return, and it suppresses the return alert for both `MBEDTLS_ERR_SSL_FATAL_ALERT_MESSAGE` and `MBEDTLS_ERR_SSL_BAD_HS_CLIENT_HELLO`. The peer goes and the server stays quiet.

One detail you may find useful: in that tree the return code is `BAD_HS_CLIENT_HELLO`. While the server sits in `MBEDTLS_SSL_CLIENT_HELLO` it parses the record header inside `ssl_parse_client_hello` (`ssl_srv.c:952`), which rejects every content type other than handshake and returns there, ahead of the alert path. Same outcome by a different route.

That is a 2023 source reading, so your WD53s remain the real evidence. The code stands as written: the alert is the standards-correct thing to send either way.

